PR #5, opened 2020-11-13
micado trimmed filters

miguelverdugo:
Hi, 
I created trimmed versions of the MICADO filters. 

Reason: MICADO filters cover (unnecessarily imho) the whole instrumental spectral range (0.8-2.5um). This increases the memory footprint, slow down calculations and raise synphot errors if spectra and filters do not fully overlap. 

The filters have been trimmed at  ``central_wavelength +/- 1.5*equivalent_width`` and visually inspected so they are not cut where they shouldn't


hugobuddel:
Some of them seem to cut-off at a rather arbitrary point, e.g. MICADO/filters/TC_filter_FeII.dat cuts off at
```
1.6790   0.0027  
```
Wouldn't it be simpler to cut them off where they reach 0.0001? E.g. extrapolating beyond the edges is easier when the edge is at 0.0001.

Also the precision is now always limited to 4 digits, but that should suffice I'd say.

miguelverdugo:
I never answered this. Sorry. I remade the MICADO filters for this and other reasons. I'm closing this pull request and opening a new one.

PR #8, opened 2021-02-16
MICADO trimmed filters

miguelverdugo:
I created trimmed versions of the MICADO filters, capped to zero at the edges. 

Reason: MICADO filters cover (unnecessarily imho) the whole instrumental spectral range (0.8-2.5um). This increases the memory footprint, slow down calculations and raise ``synphot`` errors if spectra and filters do not fully overlap.

Additionally, if filters are not capped to zero at the edges, ``synphot`` may overestimate the flux within the filter under particular circumstances.  See below for an example. 

````python

from synphot import SpectralElement, SourceSpectrum, Observation
from synphot.models import Empirical1D, BlackBodyNorm1D
import numpy as np
import astropy.units as u

# Making a square K-band filter. 
# One is capped to zero, the other at 0.0001 (MICADO-like filter)
waves  = np.arange(0.7, 2.5, 0.01) * u.um   
trans1 = np.ones(waves.shape) 
trans2 = np.ones(waves.shape)  
trans1[(waves.value<1.8) | (waves.value>2.4)] = 0
trans2[(waves.value<1.8) | (waves.value>2.4)] = 0.0001

# making bandpasses
bp1 = SpectralElement(Empirical1D, points=waves, lookup_table=trans1)
bp2 = SpectralElement(Empirical1D, points=waves, lookup_table=trans2)

# Black-body spectrum with 10C 
bb_spec = SourceSpectrum(BlackBody1D, temperature=283)

# "instrumental" magnitudes
mag1 = Observation(bb_spec, bp1).effstim(flux_unit=u.ABmag)
mag2 = Observation(bb_spec, bp2).effstim(flux_unit=u.ABmag)
print("magnitude difference:", mag1 - mag2)
#> magnitude difference: 7.225492730726543 mag


````


hugobuddel:
This seems better than #5. I've checked a few files and they al made sense.

I was wondering whether this change would cause any troubles, because now it is almost guaranteed that the curves do not perfectly overlap with each other. But if that is the case, than that should be considered a bug in that software, and not in the IRDB.

I'll merge this because KL is on leave.

/merge

hugobuddel:
Thanks for fixing this @miguelverdugo !

PR #19, opened 2021-04-16
Oc/04 slitwheel

oczoske:
I could approve this myself, but maybe someone else wants to have a quick look. I have implemented SlitWheel in `METIS.yaml`, so it is accessible to imaging and spectroscopic modes. Whether it is included or not is controlled by `!OBS.slit`. By default, it is `false` in imaging modes and set to `C` (arbitrarily) in long-slit spectroscopic modes.

### DIFF
--- a/METIS/METIS.yaml
+++ b/METIS/METIS.yaml
@@ -1,5 +1,6 @@
 ---
 ### METIS
+
 object: instrument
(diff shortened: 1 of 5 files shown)


astronomyk (approved):
Looks good

PR #21, opened 2021-11-24
edited wavemax to reflect the loss of Q band

astroleo:
edited wavemax to reflect loss of Q band


oczoske:
That's very nice. Accepted.

PR #22, opened 2021-12-07
Rescaled and partially reformatted PSF file to avoid jumps in L-band spectra

astronomyk:
# Rescaled PSF file

- By removing the K-band PSF (2.2um) the flux-jump problem in the L-Spec filter disappears. 
- The PSFs have been rescaled to the native pixel scale of the LM and N band imagers, so that simulations run faster. 
- The size of the PSF kernel images is also capped at 512 pixels (or ~3.5") as there was no need to convolve the images with PSF kernels larger than the actual detector size (N-band PSFs were effectively ~3700 x 3700 pixels). This saves space and time.
- The PSF kernels were also rescaled to sum(psf) = 1 for simplicity.

If you can think of any reason why the METIS package shouldn't use this trimmed down PSF file in the future, please comment on this issue.


rvboekel:
Hi Kieran

- cf. capping PSFs to 3.5": I think this completely fine. How large is the FoV of the the PSF files you get from the AO simulations anyway? Can you check what fraction of the flux you would "miss" from the PSF if you cut off like this? Ideally this fraction should be very small.
-  the PSF kernel should be normalised to sum up to 1 (after checking that the "capped" PSF indeed contains ~all of the flux). If a significant fraction (say: "epsilon") of the flux is missing in the capped array, then the PSF should be normalised so that it sums up to (1-epsilon).

astronomyk:
Hi Roy,

To answer your questions:

**AO Sim PSF FOV**

The size of the AO PSF FOV 

| wavelength   | FOV |
|-------|-----------|
| 3.8   | 7.8" |
| 8.65  | 17.8" |
| 11.63 | 24" |
| 12.3  | 25"  |

All apart from the L-band PSF are overkill given that our imaging FOVs are 11" and 14" for LM and N respectively.

**Missing flux:**
This begs the question of what percentages of missing flux is acceptable as "very small"? ~2%, <0.1%. A 512 pixel kernel misses on average ~2% of the flux. A 1024 pixel kernel misses <1% of the flux. The 1024 kernel is however 4x larger, and therefore the convolution process is 4x slower. 

Here are the flux percentages contained within a kernel size of 512x512 (~2.8" in LM, ~3.5" in N)

| wavelength   | sum |
|-------|-------|
| 3.8   | 0.987 |
| 8.65  | 0.983 |
| 11.63 | 0.977 |
| 12.3  | 0.975 |

Here are the flux percentages contained within a kernel size of 1024x1024 (~5.6" in LM, ~7.0" in N)

| wavelength   | sum |
|-------|-----------|
| 3.8   | 0.997 |
| 8.65  | 0.993 |
| 11.63 | 0.991 |
| 12.3  | 0.991  |

The current PSFs are scaled such that they sum to these values.

Given these values, I guess we should use the 1024 kernel then.

rvboekel:
Thanks for the quick turnaround! I think there are two considerations:

-) scientific analysis: here using the 512 pix PSFs is fine nearly all the time, we will rarely care about flux that is more than 256 pixels away from a target (1.40 arcsec for IMG-LM, 1.74 arcsec for N). Moreover, with the normalisation as you describe it, the flux levels would still be correct (so the associated flux errors are not 1-2% but essentially zero). 
-) cosmetics. If you image e.g. a field of stars with some extremely bright targets in the field, then you will see the 512x512 PSF kernel as a square box around the bright stars if you make a hard stretch on the image to see the faint stars.

In some exceptional cases (e.g. alpha Cen in the N-band) we would care about more field than 1.74 arcsec radius (I think the AO control radius in the N2 filter is a bit larger than this. But that is a rather special application.

Are the kernels calculated/interpolated on the fly? If it is easy to do, one could make the 512 kernel the default  (which will work fine in the majority of cases), but allow the user to indicate that a 1024 kernel should be used at the expense of computing time if the particular case requires that. It is fine to have such "advanced" functionality, as long as the default mode is fine for the majority of cases so that the "ignorant" user will do the right thing automatically. Users with "extreme" applications can be expected to be extra careful.

METIS/tests/rescale_psf.py
```diff
@@ -0,0 +1,50 @@
...
+from scipy.ndimage import zoom
+from astropy.io import fits
+
+
+def rescale_image(old_hdu, new_cdelt, max_size=512):
+    scale_factor = old_hdu.header["CDELT1"] / new_cdelt
+    image = zoom(old_hdu.data, scale_factor)
+    image /= image.sum()
+    y, x = np.divmod(np.argmax(image), image.shape[1])
+    r = max_size // 2
```
oczoske, line 11: Using `np.unravel_index` might be more explicit:
```
y, x = np.unravel_index(np.argmax(image), image.shape)
```
Code would be a bit less cryptic if you used longer variabel names...
I'm slightly uneasy about using argmax for the centre - it seems unlikely, but there might be cases where we're dealing with wobbly PSFs (well, the HCI PSFs do wobble).

oczoske:
Bear in mind that the irdb/METIS includes just one example PSF file (and I picked it fairly randomly, if I remember correctly), so all changes to the PSF must be done on the fly by Scopesim.  
In LSS-L, the PSF is 45% larger at the red end (4.2um) than at the blue end (2.9um). Ideally, the PSF would be rescaled for each wavelength layer. I haven't tried it, but I fear that would take a long time...

METIS/tests/rescale_psf.py
astronomyk: Yep. Removing the K-band PSF is why there is no more jump. The cause remains a mystery.

Re wobbly PSFs. I'm happy to use a different method to determine the centre. This was just a super quick hack to get the current PSF reformatted so that we could include it in the release, purely due to the time constraints,

astronomyk:
@rvboekel - the square borders of a super bright star are indeed an issue with an "undersized" PSF image. I have a method to remove these, but it's not 100% perfect.

@oczoske @rvboekel - I had not considered the fact that the PSF is almost 50% wider at the red end of L-spec. The new PSF file would mean that there is only one constant-width PSF over the L-band (and M-band) spectral wavelength range. The N-band spectra will still use 3 different PSFs - and yes there is still a small discontinuity seen in the spectra.

An alternative to this might be to generate the PSFs on the fly (**~2 sec per PSF**) using the [AnisoCADO ](https://anisocado.readthedocs.io/en/latest/)package (and the corresponding ScopeSim effect: `AnisocadoPSF`) which can be altered to generate a given number of PSFs within the observed wavelength range (e.g. 2.9, 3.2, 3.5, 3.8, 4.1um). There are several added bonuses that come with generating the PSF on the fly:

The AnisocadoPSF can be generated for:
- different turbulence profiles (EsoQ1, EsoMedian, EsoQ4)
- user defined Airmass and Seeing conditions
- off axis guide stars (although I guess METIS will not be to interested in this)
- different pupil alignments relative to the spider orientation
- N missing ELT M1 mirror segments

Obviously these are mostly options for the "power user", however the Airmass and Seeing constraints may be of interest to the basic user.

The PSFs generated by AnisoCADO obviously don't match up perfectly with the current SCAO PSF, however AnisoCADO is a generic SCAO PSF simulator (i.e. not MICADO specific, regardless of name) and therefore shouldn't be too far off what we expect for METIS.

The image below shows a comparison of the PSFs generated. Note the colour bar scales. The major differences are on the <10-5 level. The central areas however look extremely similar, and their flux distributions in the central 64 pixels also match up very well (<5% difference). The Airy rings are however less pronounced in the AnisoCADO image due to it using a hexagonal segmented mirror pupil image.

image

rvboekel:
Hi @astronomyk,

Thanks for the update. Having a single PSF applied to all wavelengths in a spectrum I think unfortunately is a no-go for the longslit modes. For the LMS it is probably OK, since the simultaneous spectral range we fit onto one detector is very small so the PSF(lambda) dependence, while certainly measurable on high SNR sources in a single setting, is not impacting the science. But in the longslit modes, the relative bandwidth (range(lambda)/mean(lambda)) is big, in the N-band the fwhm at the red extreme of the wavelength range is ~75% larger than ad the blue extreme. This is much to big to be ignored, moreover because much of the scientific "action" for METIS will be on the range of scales from "we barely resolve it" to "our apparent source size is a single digit number of resolution elements". The "error" in the PSF size currently cuts right into that.

So we will have to find a compromise between realism and computational demands. In my simple calculation, I convolve with a much smaller kernel, only a a few x lambda/D in radius (corresponding to 64x64 pixels, but I use an over-sampling of 2 at this point, so it corresponds to 32x32 physical detector pixels. The result is re-binned to the actual detector pixels after the convolution). I simply loop this over the wavelength grid, i.e. calculate my analytical PSF and convolve the model flux for that particular "pixel in dispersion direction" (PiDD) with the PSF for each individual PiDD, i.e. the process is repeated 2000 times (I assume 2000 useful pixels in dispersion direction, you have a better estimate from the actual design). This procedure requires pretty much dead on 2 seconds, so that is workable - BUT: I don't simulate the whole slit, my "simulator" is just an add-on of the radiometry code and does only point sources, so in addition to the small Kernel, it also only simulates a spatial region of 32 (physical) pixels in the spatial direction, i.e. ~0.22 arcsec. A proper simulator needs to do more than that; in the simplest case the whole slit, that gets expensive. One could adopt a size of 256 (physical) pixels as a default, this corresponds to an angular size of 1.74 arcsec for the N-band and 1.40 arcsec for the L/M bands. If the apparent size of the source is smaller than that (which will be the case in almost all applications), that would be sufficient. The size of the PSF kernel does not necessarily need to be the same, it can be smaller.

Nice PSF plots! I would nevertheless advocate, at least for now, to not calculate PSFs on the fly, but rather to have a pre-calculated PSF (cube), wavelength sampling TBD, in which to interpolate. This will probably be much faster.

The question then is what the wavelength sampling should be for the PSFs. In my simple calculations, as said, I just do it for every physical (on the detector) wavelength bin, so the results are perfectly smooth (see attached simulation as an example). But that will probably take too long and also is more than needed for smooth results. This will require some experimenting to find a good compromise between speed and smoothness (substantial "jumps" in the PSF width are not desired, since doing precise analysis of the profiles vs. wavelength will be among the prime applications, think spectra-astrometry). 

cheers,
Roy

[longslit_Nlow_Sirius_scaled50.0000Jy.fits.gz](https://github.com/AstarVienna/irdb/files/7676929/longslit_Nlow_Sirius_scaled50.0000Jy.fits.gz)

rvboekel:
Just noticed that the "sirius scaled to 50 Jy" simulation was an old one from the "Aquarius" days, here's one for Vega (single DIT) that we did last week.

[Vega_longslit_N_slit_D_Nov2021.fits.gz](https://github.com/AstarVienna/irdb/files/7677136/Vega_longslit_N_slit_D_Nov2021.fits.gz)

oczoske:
I suggest closing (aka rejecting) this request. For spectroscopy, the PSF is scaled quasi-continuously through the FOV cube, removing any discontinuities. The irdb also includes single-extension PSF-files for spectroscopy, so the FOV is not split anymore. However, this is just for one AO situation, and the other PSF files on Markus Feldt's website cannot be changed by us, of course.

rvboekel:
Hi both,

cf. PSFs, I think the long-exposure PSFs recently sent by Olivier Absil are currently our best guesses for the actual performance. The old ones from Markus Feld’s website are highly idealised and outdated.

cheers,
Roy

PR #23, opened 2021-12-09
Implement AutoExposure, some cleaning

oczoske:
Apply effect `AutoExposure` in METIS detectors. To test you'll need branch `oc/autodit` of Scopesim.

### DIFF
--- a/METIS/METIS_DET_IFU.yaml
+++ b/METIS/METIS_DET_IFU.yaml
@@ -9,6 +9,9 @@ properties:
     temperature: -233
     dit: "!OBS.dit"
     ndit: "!OBS.ndit"
+    mindit: 0.04              # seconds, E-TNT-MPIA-1004, v1-0
+    full_well: !!float 1.E5   # electrons, E-TNT-MPIA-1004, v1-0
+
 
 effects:
     - name: detector_array_list
@@ -23,12 +26,23 @@ effects:
       kwargs:
           filename: QE_detector_H2RG_METIS.dat
 
+    - name: auto_exposure
+      description: automatic determination of DIT and NDIT
+      class: AutoExposure
+      include: True
+      kwargs:
+        fill_frac: "!OBS.auto_exposure.fill_frac"
+        exptime: "!OBS.auto_exposure.exptime"
+
+    - name: exposure_action
+      description: Summing up sky signal for all DITs and NDITs
+      class: SummedExposure
+
     - name: dark_current
       description: METIS LMS dark current
       class: DarkCurrent
-      # [e-/s] level of dark current for each detector
       kwargs:
-          value: 0.1
+          value: 0.1  # [e-/s]
 
     - name: detector_linearity
       description: Linearity characteristics of H2RG chips
--- a/METIS/METIS_DET_IMG_LM.yaml
+++ b/METIS/METIS_DET_IMG_LM.yaml
@@ -10,10 +10,8 @@ properties:
   temperature: -230
   dit: "!OBS.dit"
   ndit: "!OBS.ndit"
-  width: 256
-  height: 256
-  x: 0
-  y: 0
+  mindit: 0.04              # seconds, E-TNT-MPIA-1004, v1-0
+  full_well: !!float 1.E5   # electrons, E-TNT-MPIA-1004, v1-0
 
 effects:
     - name: detector_array_list
@@ -23,24 +21,19 @@ effects:
       kwargs:
         filename: FPA_metis_img_lm_layout.dat
 
-    - name: detector_window
-      description: detector window
-      class: DetectorWindow
-      include: False
-      kwargs:
-        image_plane_id: 0
-        pixel_size: 0.018
-        x: "!DET.x"
-        y: "!DET.y"
-        width: "!DET.width"
-        height: "!DET.height"
-        units: pixel
-
     - name: qe_curve
       description: Quantum efficiency curves for each detector
       class: QuantumEfficiencyCurve
       kwargs:
-          filename: QE_detector_H2RG_METIS.dat
+        filename: QE_detector_H2RG_METIS.dat
+
+    - name: auto_exposure
+      description: automatic determination of DIT and NDIT
+      class: AutoExposure
+      include: True
+      kwargs:
+        fill_frac: "!OBS.auto_exposure.fill_frac"
+        exptime: "!OBS.auto_exposure.exptime"
 
     - name: exposure_action
       description: Summing up sky signal for all DITs and NDITs
@@ -49,9 +42,8 @@ effects:
     - name: dark_current
       description: METIS LM dark current
       class: DarkCurrent
-      # [e-/s] level of dark current for each detector
       kwargs:
-          value: 0.1
+          value: 0.1   # [e-/s]
 
     - name: detector_linearity
       description: Linearity characteristics of H2RG_1-5 chips
--- a/METIS/METIS_DET_IMG_N_Aquarius.yaml
+++ b/METIS/METIS_DET_IMG_N_Aquarius.yaml
@@ -6,9 +6,12 @@ name: metis_detector_array
 description: Aquarius detector for METIS IMG-NQ
 
 properties:
-    temperature: -265
-    dit: "!OBS.dit"
-    ndit: "!OBS.ndit"
+  image_plane_id: 0
+  temperature: -265
+  dit: "!OBS.dit"
+  ndit: "!OBS.ndit"
+  mindit: 0.008            # seconds, E-REP-NOVA-MET-1191, v2-0
+  full_well: !!float 1.e7  # electrons, E-REP-NOVA-MET-1191, v2-0
 
 effects:
     - name: detector_array_list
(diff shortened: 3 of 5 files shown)


METIS/METIS_DET_IMG_N_GeoSnap.yaml
```diff
@@ -25,16 +27,23 @@ effects:
...
       description: Summing up sky signal for all DITs and NDITs
       class: SummedExposure
 
     - name: dark_current
       description: METIS GeoSnap dark current
       class: DarkCurrent
-      # [e-/s] level of dark current for each detector
       kwargs:
-          value: 100.
+          value: 1.e5   # [e-/s], E-REP-NOVA-MET-1191, v2-0
```
astronomyk, line 46: `value : !!float 1.e5`

Unless there is new scientific notation detection in the newest pyyaml version. (Wouldn't suprise me if there is)

METIS/default.yaml
```diff
@@ -24,6 +24,9 @@ properties:
   pupil_angle: 0
   dit: 1
   ndit: 1
+  auto_exposure:
+    exptime: None     # seconds, overrides !OBS.dit and !OBS.ndit
+    fill_frac: 0.75   # fraction to which full well is filled in one DIT
```
astronomyk, line 29: Theoretically this could be >1 right? Then it would be possible to also use auto exposure for frames where it is acceptable for the brightest source to be saturated, or?

astronomyk (commented):
Geosanp YAML dark current is issing the `!!float` prefix
Otherwise looks fine

METIS/METIS_DET_IMG_N_GeoSnap.yaml
oczoske: Well spotted! Fixed.

astronomyk (approved):
Lightning fast response times! I approve ;) 👍

METIS/default.yaml
oczoske: That's right, the effect does test for `fill_frac` to be less than 1. I think we can rely on the user to know what they're doing. A test for non-negativity might be justified, though.

METIS/METIS_DET_IFU.yaml
```diff
@@ -9,6 +9,9 @@ properties:
     temperature: -233
     dit: "!OBS.dit"
     ndit: "!OBS.ndit"
+    mindit: 0.04              # seconds, E-TNT-MPIA-1004, v1-0
```
rvboekel, line 12: For the LMS detectors, the fast readout mode is currently not baselined. Therefore, min_dit = 1.3 seconds (even though these are physically the same detectors as the one in the IMG-LM, which we run at up to 25 Hz full frame. Reason: fast readout is more noisy and gives much higher power (thermal dissipation) at the detector).

METIS/METIS_DET_IFU.yaml
```diff
@@ -23,12 +26,23 @@ effects:
...
+      description: Summing up sky signal for all DITs and NDITs
+      class: SummedExposure
+
     - name: dark_current
       description: METIS LMS dark current
       class: DarkCurrent
-      # [e-/s] level of dark current for each detector
       kwargs:
-          value: 0.1
+          value: 0.1  # [e-/s]
```
rvboekel, line 45: if you're really really really strict on units it would even be [e-/s/pix]
("grammar-nazi" equivalent for units, I know ....)

METIS/METIS_DET_IMG_LM.yaml
```diff
@@ -10,10 +10,8 @@ properties:
   temperature: -230
   dit: "!OBS.dit"
   ndit: "!OBS.ndit"
-  width: 256
```
rvboekel, line 13: do these numbers correspond to the dimensions of a convolution kernel? If they refer to the detector, they should be 2048, no?

METIS/METIS_DET_IMG_N_Aquarius.yaml
```diff
@@ -6,9 +6,12 @@ name: metis_detector_array
...
 properties:
-    temperature: -265
-    dit: "!OBS.dit"
-    ndit: "!OBS.ndit"
+  image_plane_id: 0
+  temperature: -265
+  dit: "!OBS.dit"
+  ndit: "!OBS.ndit"
+  mindit: 0.008            # seconds, E-REP-NOVA-MET-1191, v2-0
+  full_well: !!float 1.e7  # electrons, E-REP-NOVA-MET-1191, v2-0
```
rvboekel, line 14: For the Aquarius, we had 2 full wells (for the GeoSnap as well actually), one for the high capacity mode (in Aquarius lingo called "low gain") of ~1e7 [e-/pix] and one for the low capacity mode ("high gain") of 1e6 [e-/pix]. We were planning to use the low capacity mode generally, since operating at illumination levels requiring the high capacity mode, the Aquarius suffered badly from ELFN (Excess Low Frequency Noise), killing the sensitivity.

METIS/METIS_DET_IMG_N_GeoSnap.yaml
```diff
@@ -9,6 +9,8 @@ properties:
     temperature: -240
     dit: "!OBS.dit"
     ndit: "!OBS.ndit"
+    mindit: 0.011             # seconds, E-REP-NOVA-MET-1191, v2-0
+    full_well: !!float 2.8e6  # electrons, E-REP-NOVA-MET-1191, v2-0
```
rvboekel, line 13: For the GeoSnap we also have a high capacity mode and a low capacity mode; see our hackathon minutes. [https://docs.google.com/document/d/1XLClrK8YGed4pQHLds3Q0T-LQHBLf4K4uW-mVcvR-yM/edit#heading=h.5a4s7xsi6p5y](url)

METIS/default.yaml
rvboekel: I think we should try to make ScopeSim assist the user as much as possible in "being sensible" here: 

1. test whether the peak intensity [e-/s/pix] times the DIT [s] is more than the full well x the allowable fill fraction. 
2. If this is the case, all is good
3. If not, then:
    - if the linearity option is on, continue but put out a prominent warning message that the source saturates (goes into the non-linear regime) so that photometry will be unreliable (perhaps indicate which pixels are affected?)
    - if the linearity option is off, refuse to continue, put out an error message saying that the source saturates, and that in order to get output at all, the linearity option needs to be switched on (say how, and stress that this will still give unreliable photometry). But really, that the user should consider changing the setup (e.g. use an ND filter, or use a coronagraphic observation; cf. discussions with Liege coupling ScopeSim with HEEPS).

rvboekel (commented):
I left some comments here and there. Is that the proper way to do the "review"?
thanks!
Roy

METIS/default.yaml
oczoske: 1, 2, and 3a are already implemented in Scopesim. The code tests whether the automatically determined DIT is less than MINDIT. If it is, then DIT is set to MINDIT and the warning is issued. Indicating the affected pixels would be require non-negligible coding work and is something that seems quite alien to Scopesim. As for 3b, this seems rather intrusive - it is not obvious how the user would get around that blockage should they wish to do so. It seems preferable to leave the (non-)linearity on by default. It is then an effort for the user to switch it off, and we assume they only do it for a reason.

METIS/METIS_DET_IFU.yaml
oczoske: Changed to 1.3 s (already in branch `dev_master`)

METIS/METIS_DET_IMG_LM.yaml
oczoske: Red means it was removed in the commit, so `width` and the rest do not exist any more. They were related to the `DetectorWIndow` effect, which enables simulating just a fraction of the detector. I removed that effect as the detector is small anyway (it is useful for MICADO, for instance). It can always be reinstated.

METIS/METIS_DET_IMG_N_Aquarius.yaml
oczoske: Changed to 1e6.

METIS/METIS_DET_IMG_N_GeoSnap.yaml
oczoske: Is it documented anywhere which detector mode is to be used for which observing mode? Do we expect a one-to-one relation? Not sure how to implement this - simplest would be to have to yaml files, one for high-, one for low-capacity mode. That would cause redundancies in the other parameters, though.

METIS/default.yaml
rvboekel: Thanks! Switching the linearity on by default, with a warning if a simulation goes into the non-linear regime, and requiring special action if an expert user wants to circumvent that for test purposes, makes sense and is certainly more elegant. Perhaps we should always include the "range of where we are on the linearity curve" with any simulation (like: lowest flux level we are at 99.x1 percent linearity, highest flux level we are at 98.x2 percent). Perhaps have a text file with "details" that is created then running a simulation (or packed into the fits header); we could put information in there that in most cases the user won't care too much about, but for users getting into the nitty-gritty or trying to do high accuracy applications may/should care about). Something for future consideration perhaps.

METIS/METIS_DET_IMG_N_GeoSnap.yaml
rvboekel: No, there is no "final" concept for this yet. What I have done for now in the simulations is simply to do a check of the signal levels [e-/s/pix], and if they are lower than some fraction of the "detection capacity" (defined as well x frame rate, units [e-/s/pix]) then I assume the LCM is used, if the signal level supersedes said limit then HCM is used. In reality I think LCM will always be used for spectroscopy (but perhaps we need to allow HCM, too, in order to be able to observe very bright sources, TBC). And for imaging typically HCM will be used, for broadband imaging always for sure, though for the narrow band filters it may be beneficial to use LCM (TBC) but it depends on the observing conditions, what the real background will be, and on the science application. So here there is no simple answer unfortunately.

Issue #28, opened 2022-01-28
Ensure Focal Plane Array (FPA) parameters are up to date


hugobuddel:
Ric created a [technical note about some the Focal Plane Orientation](https://bscw.mpe.mpg.de/bscw/bscw.cgi/5161261?op=preview) (ELT-TRE-MCD-56300-0132, v1.3, dated May 6th 202), that seems to not be incorporated in [MICADO/FPA_array_layout.dat](MICADO/FPA_array_layout.dat), which is from October 2020.

It is not entirely clear what needs to change (if anything indeed needs to change).

@astronomyk can you verify that the MICADO package is up to date with the above note?

astronomyk:
Thanks for reminding me of the document. I haven't read it, so I can not
confirm if the ScopeSim Micado package is up to date. I'll add this to my
to-do list for when I'm back at work.



Hugo Buddelmeijer ***@***.***> schrieb am Fr., 28. Jän. 2022,
14:58:

hugobuddel:
Closed by #32

PR #32, opened 2022-02-09
Kl/micado detector gap update 2

astronomyk:
Updates the big/small detector gaps to 6.5/2.5mm and places the big detector gap between the central (Nr 5) and middle-right detector (Nr 4). Addresses issue #28


hugobuddel:
Thank you Kieran!

PR #37, opened 2022-02-21
oc/cleaner compile

oczoske:
The published packages should not contain crap like `__pycache__`, `.ipynb_checkpoints` and the like. As `shutil.make_archive` does not have the option to exclude stuff `shutil.copytree` is used to  copy the package to a temporary directory, ignoring a list of `ignore_patterns` before archiving. 
Also some pylint cosmetics. `compile` is a builtin and should not be used as a variable, changed to `compile_pkg` (and `upload_pkg` for symmetry). This does not affect the user interface.


oczoske:
Sorry, this has too much stuff in it.

PR #38, opened 2022-02-21
Publish should not include __pycache__ and hidden files

oczoske:
The published packages should not contain crap like `__pycache__`, `.ipynb_checkpoints` and the like. As `shutil.make_archive` does not have the option to exclude stuff `shutil.copytree` is used to copy the package to a temporary directory, ignoring a list of `ignore_patterns` before archiving.
Also some pylint cosmetics. `compile` is a builtin and should not be used as a variable, changed to `compile_pkg` (and `upload_pkg` for symmetry). This does not affect the user interface.

### DIFF
--- a/irdb/publish.py
+++ b/irdb/publish.py
@@ -1,7 +1,9 @@
+"""Publish and upload irdb packages"""
 import sys
 import os
 from os import path as pth
 import shutil
+from tempfile import TemporaryDirectory
 from datetime import datetime as dt
 import yaml
 import pysftp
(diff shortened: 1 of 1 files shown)


irdb/publish.py
```diff
@@ -64,21 +88,37 @@ def rename_package(pkg_path):
...
-    pkg_dir = pth.join(PKGS_DIR, pkg_name)
-    # new_pkg_path = shutil.make_archive(pkg_name, "zip", pkg_dir, pkg_name)
-    new_pkg_path = shutil.make_archive(pkg_name, "zip", PKGS_DIR, pkg_name)
+    """
+    Create a zip file of packages in `pkg_names`
+
+    Directories `__pycache__` and hidden files (starting with `.`) are
+    ignored.
+    """
+    ignore_patterns = shutil.ignore_patterns("__pycache__", ".*")
```
AstarVienna, line 107: Should we also include docs here? Or should docs (e.g. ipynb files) also be compiled into the package?

astronomyk: Btw, thanks for doing this. It's been on my to-do list for the last week, as I wasn't sure about whether we should include the docs folders in the compiled packages

astronomyk (approved):

irdb/publish.py
oczoske: Notebooks should definitely be in the compiled package - on the website we will tell users how to download packages and where to find the notebooks there. OTOH readthedocs also renders the notebooks, right?

astronomyk:
Yep, RTD renders the notebooks and runs all code cells automatically. We
just have to add the filename of the notebook to the METIS/docs/readme.rst
file


-------------------------------------
Kieran Leschinski
Department of Astrophysics
University of Vienna


Am Di., 22. Feb. 2022 um 09:37 Uhr schrieb oczoske ***@***.***

PR #43, opened 2022-03-25
 Rename ELT.telescope_reflection to just ELT.reflection pt2

hugobuddel:
Same as #42  but then to the master branch. The branch structure of the IRDB is not clear to me.

It is clear that `telescope_reflection` is the reflection of the telescope, because it is in a telescope object, so it is nicer to call it simply `reflection`.

(Ulterior motive: MicadoWISE models telescope objects with a Telescope Python class. It is much nicer to have a `Telescope.reflection` attribute than a `Telescope.telescope_reflection` attribute. These attributes automatically get converted to hierarchical FITS headers, taking the first three letters of the object, the effect, and then the kwarg. `TEL TEL FILENAME` is ugly, `TEL REF FILENAME` is a bit better.)

(I also plan to remove the `micado_` from many of the effect names in MICADO, because it is obvious they are for MICADO. But those don't require much consensus.)

@oczoske the `telescope_reflection` name is from you. Is it okay to change it to `reflection`?

### DIFF
--- a/ELT/ELT.yaml
+++ b/ELT/ELT.yaml
@@ -14,7 +14,7 @@ properties :
     filename : "TER_ELT_5_mirror.dat"
 
 effects :
-  - name : telescope_reflection
+  - name : reflection
     description : single combined reflection curve (ESO-333023)
     class : SurfaceList
     include : True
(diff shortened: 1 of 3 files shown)


hugobuddel:
This one also updates the METIS tests, which do not exist in the `dev_master` branch.

hugobuddel:
OK, perhaps not change this now. I'll ensure it works well in MicadoWISE anyway.

However, why is just `reflection` not specific enough, since it is already in the `telescope` namespace so to say? Are you in some way referring directly to `telescope_reflection` without actually going through the ELT / telescope object?

As in, I might be missing a usecase that I'm therefor not considering. All my code is like, 1) get the telescope yaml/class/whatever, 2) of this telescope thing, get the telescope_reflection. From that perspective `telescope_` prefix seems to be distracting repetition.

PR #45, opened 2022-03-25
Fix all MICADO tests

hugobuddel:
This should fix all MICADO tests. Or well, `test_background_is_within_2x_of_eso_etc` is simply marked `xfail` and replaced by `test_background_is_within_4x_of_eso_etc`

This includes #44 , so maybe merge that one first

Closes #16

Issue #12, opened 2021-03-20
Add CI to at least have astropy read in all tables?


hugobuddel:
Perhaps we can add some continuous integration tests that read in all the tables?

That would have caught #10 right of the bat!

oczoske:
I have added such tests in [test_metis.py](https://github.com/astronomyk/irdb/blob/master/METIS/tests/test_metis.py) (yamls, filters, ter curves, fpa layout, linearity). There's quite a bit of redundancy between the file types but I wanted to keep them separate. Maybe you want to adapt that to MICADO?  Suggestions for improvements are welcome.

hugobuddel:
Thanks I'll look into that.

It seems that it are the MICADO tests that prevent the CI from succeeding, so fixing those would probably already be useful.

I might be able to look into that next month; I didn't really have these tests on my radar and now it bites me.

PR #47, opened 2022-03-27
Fix / xfail all tests

hugobuddel:
This ensures all tests in the IRDB either pass or xfail.

Requires a fix to ScopeSim and ScopeSIM_Templates as well:

- ScopeSIM to fix reading ELT tables: https://github.com/AstarVienna/ScopeSim/pull/121
- ScopeSIM_Templates to fix cluster units: https://github.com/AstarVienna/ScopeSim_Templates/pull/25

@astronomyk @oczoske I propose that we merge these three PR's and then subsequently keep everything green here.

This PR encompasses #45 and #46 (it is an extension of the same branch), so we can either merge those first, or just directly merge this.

### DIFF
--- a/MICADO_Sci/MICADO_Sci_SPEC.yaml
+++ b/MICADO_Sci/MICADO_Sci_SPEC.yaml
@@ -50,6 +50,9 @@ effects:
     kwargs :
         filename : "TRACE_SCI_3arcsec.fits"
         center_on_wave_mid: True
+        x_colname : "x0"
+        y_colname : "y0"
+        s_colname : "s0"
 
 ---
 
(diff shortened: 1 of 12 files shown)


MICADO_Sci/MICADO_Sci_SPEC.yaml
```diff
@@ -50,6 +50,9 @@ effects:
     kwargs :
         filename : "TRACE_SCI_3arcsec.fits"
         center_on_wave_mid: True
+        x_colname : "x0"
+        y_colname : "y0"
+        s_colname : "s0"
```
hugobuddel, line 55: This is not a proper fix, instead the trace files themselves should be updated to the new format.

hugobuddel: Created #48 for updating these trace files, so this can be merged.

hugobuddel:
There are still some failed tests, but I'll merge this anyway, so we make some progress.

* 2 tests fail because the off-by-one error for large simulations, this is fixed in scopesim dev_master, but not yet in master, will be fixed automatically.
* 1 test fails because it cannot read the file, those need to be tested. Issue #12 is there to do that.

PR #50, opened 2022-04-04
Fix #-strings in IRDB

hugobuddel:


hugobuddel:
Replaced by #51 because this one also contains fix for #-strings which is only in dev_master , so this one should be merged into irdb dev_master

astronomyk (approved):

PR #54, opened 2022-04-05
Draft: Test to only count failure if the ScopeSIM branch is the same.

hugobuddel:
DO NOT MERGE AS-IS

The goal of this PR is that, if you have a feature branch that you want to merge into dev_master, that then you want your tests to run against ScopeSIM dev_master. Those tests should all pass. However, it is very useful to know whether your tests would also pass against ScopeSIM master, but you don't want that to cause the CI to fail; you just want to know whether your new code works against the ScopeSIM master branch.

This MR ensures that you only get failing red checks when the test fails against the same branch of ScopeSIM. It will also run the tests against the ScopeSIM master branch, but those will always succeed when ran in a PR. However, when you run the CI outside of a PR, then all tests can become red.

What this does, is that if there is a bug, then
- The bug is ignored if we are testing IRDB dev_master against ScopeSIM master (or vice-versa) in a PR: the 3 greens in https://github.com/AstarVienna/irdb/runs/5840214584?check_suite_focus=true
- The bug in NOT ignored if we are testing against the same branch in both repositories, e.g. dev_master: the 3 reds in https://github.com/AstarVienna/irdb/runs/5840214584?check_suite_focus=true
- The CI is triggered outside a PR, e.g. manually, because that is an indication that you are interested in all problems: the 6 reds in https://github.com/AstarVienna/irdb/actions/runs/2098478168

NOTE: For ease of testing, this PR only runs pytest on one test

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,11 +43,11 @@ jobs:
             echo "Target IRDB branch: ${GITHUB_BASE_REF}"
             echo "This IRDB branch:   ${GITHUB_REF_NAME}"
             echo "ScopeSIM branch:    ${{ matrix.scopesim-branch }}"
-            this_branch_is_scopesim_branch=$(expr "${GITHUB_REF_NAME}O" == "${{ matrix.scopesim-branch }}O" )
+            this_branch_is_scopesim_branch=$(expr "${GITHUB_REF_NAME}O" == "${{ matrix.scopesim-branch }}O" || true )
             echo "this_branch_is_scopesim_branch: ${this_branch_is_scopesim_branch}"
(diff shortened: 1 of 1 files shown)


hugobuddel:
@astronomyk can you NOT MERGE THIS, but see whether this makes sense at all? In particular

- Do you (think you) understand what I'm trying to achieve?
- Do you agree with that goal?
- Do you understand how the code works?

The code has become much more complicated than expected, so I think it would be good to explore other options to achieve this.

hugobuddel:
I'm closing this, because it is way to complicated. New plan:

- For PR's, run PR's that merge into dev_master (almost all of them) against ScopeSIM dev_master, run PR's that merge into master (normally only when merging dev_master into master) against master
- For manual test runs, run against both master and dev_master

PR #61, opened 2022-04-11
Draft: Add MICADO headers

hugobuddel:
Add initial headers for MICADO, still draft.


hugobuddel:
See also #62 because I think it would be nicer if all these have defaults. That is why these are not yet merged directly into dev_micado, because now they would break all examples. (Most of them are already broken due to https://github.com/AstarVienna/ScopeSim/issues/136 though..)

PR #64, opened 2022-04-12
Second attempt to run tests against ScopeSIM master and/or dev_master

hugobuddel:
Run tests against dev_master and/or master


hugobuddel:
I'm going to merge this, because hopefully seeing green ticks will prevent things from breaking

PR #66, opened 2022-04-13
Run notebooks in the CI

hugobuddel:
Run notebooks in the CI


hugobuddel:
@astronomyk running the notebooks in the CI! Should run automatically on the master and dev_master branches (so only after merging), but not for PRs, or when ran manually: https://github.com/AstarVienna/irdb/runs/6003188886?check_suite_focus=true

Currently only for the MICADO notebooks, and only against ScopeSIM dev_master, but it is a start.

PR #67, opened 2022-04-13
[METIS] added 2 skycalc override files for LSS and LMS, plus the yaml entries

astronomyk:
As the title says.

Consequence is that by default LSS and LMS will run with the local skycalc sky background. It the user is interested in different settings, they will either need to download the skycalc fits file for their needs, or set the ATMO overrides in METIS_LSS.yaml accordingly: 

```
---
alias: ATMO
name: LMS_ATMO_overrides
description: Setting the LMS mode to override the skycalc download

properties:
  skycalc:
    use_local_file: SKYCALC_LSS_override.fits
    # use_local_file: False    # to go back to using the skycalc server
```


Armazones/Armazones.yaml
```diff
@@ -8,32 +8,36 @@ changes:
...
+  pwv :           2.5         # [mm]
+  airmass :       "!OBS.airmass"
+  pupil_angle :   "!OBS.pupil_angle"
+  pixel_scale :   "!INST.pixel_scale"
+  background :
+    filter_name : Ks
+    value: 13.6
+    unit: mag
+  spectrum :
+    filename: "TER_armazones_default_FULL_IMG.dat"
```
oczoske, line 30: Do we still need this?

oczoske:
If I understand correctly, the default is now to use the local files provided with the irdb. In this case, they should not be called "override" but "default". I don't fully understand how this will work in practice. How does the user now tell scopesim to use skycalc rather than the default file? How do they change the file name? Ideally, this would go via `cmd['!ATMO.filename'] = ...` - ah, it's the rather wordy `!ATMO.skycalc.use_local_file`. Are the conditions for the default files well defined?  `SKYCALC_LSS_override.fits` has `TEL.SITE.HEIGHT=2640`, so it's for Paranal, not Armazones (the same is probably true for the skycalc downloads?). Does `!OBS.airmass` have an influence? That would seem necessary for simulating longer time series.

METIS/METIS_LMS.yaml
```diff
@@ -8,7 +8,11 @@ date_modified: 2022-05-12
 changes:
   - 2022-03-04 (OC) introduce decouple_detector_from_sky_headers
   - 2022-03-22 (OC) add METISLMSImageSlicer effect
+<<<<<<< HEAD
   - 2022-04-13 (KL) added ATMO overrides for skycalc_use_local_file for ScopeSim v0.5 upgrade
+=======
+  - 2022-05-12 (OC) add fp2_platescale
+>>>>>>> dev_master
```
hugobuddel, line 15: It seems you actually committed the conflicts! I mean, that seems to be what the commit message says, but probably not what you intended
```suggestion
  - 2022-04-13 (KL) added ATMO overrides for skycalc_use_local_file for ScopeSim v0.5 upgrade
  - 2022-05-12 (OC) add fp2_platescale
```

astronomyk:
Thanks for picking this up! I was actually just using github to store my
local branches as I needed to re-install my local operating system. I was
planning on sorting things out (i.e. whether I still need this branch)
after my system is running again :)

Hugo Buddelmeijer ***@***.***> schrieb am Di., 26. Juli 2022,
08:28:

PR #68, opened 2022-04-13
Add shifted cluster

hugobuddel:
Add shifted cluster


hugobuddel:
Since @astronomyk approved this, I'm going to merge it. @yixiancao do you think the code in this notebook would be useful to simulate the dithered observations? It requires doing the shifts manually, including setting the RA/Dec and the TPL.EXPNO; which is acceptable for now I think.

PR #72, opened 2022-04-19
De-spaghettified MAAT branch, pull into dev_master

astronomyk:


astronomyk:
All files are MAAT specific, so no dramas with breaking MICADO (unless the previous merge broke it)

Seems the OSIRIS tests fail though 😢

PR #85, opened 2022-09-23
Update Master for compiliing new MICADO package

astronomyk:
Things added in this PR
- MAAT, all updates that have previously been kept to the dev_maat branch.
- new E/VLT instruments. Skeleton folders as placeholders to remind me to fill these at some point
- MICADO, new keywords for the FITS headers
- Deleted SimpleCADO and MICADO_ETC packages, as these were barely/never used

### DIFF
--- a/GTC/reflectivity_GTC.dat
+++ b/GTC/reflectivity_GTC.dat
@@ -1,6 +1,6 @@
 # description : GTC telescope throughput vs. wavelength
-# author : R
-# source :
+# author : Kieran Leschinski
+# source : Made it up
 # date_created : 2022-02-17
 # date_modified : 2022-02-17
 # type : TERCurve
@@ -16,6 +16,6 @@
 # changes :
 #   - 2022-02-17 (KL) Created file
 #
-wavelength	reflection	emissivity
-     0.300	         1           0
-     0.100	         1           0
+wavelength  reflection  emissivity
+0.3         1           0
+1.0         1           0


hugobuddel:
The CI is failing for the last 3 weeks because
```
Traceback:
/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
irdb/tests/test_publish.py:5: in <module>
    from .. import publish as pub
irdb/__init__.py:12: in __getattr__
    raise ImportError(f"{name} not in irdb.PKGS dictionary")
E   ImportError: publish not in irdb.PKGS dictionary
```
I propose that should be fixed before merging into master. Your call though.

astronomyk:
CI is now failing because of a line in one of the ipynb files.
I will accept this merge and then look into the notebook.

hugobuddel:
It is probably not the notebook that is the problem, but the SourceDescriptionFitsKeywords.

The point of the tests is to prevent releasing broken code (or broken docs). This particular bug is probably not in the IRDB itself though, so maybe merging this didn't actually introduce any new problems. Nevertheless, keeping these tests green would have highlighted the ScopeSIM problem (if the issue is indeed there) before that was merged into master.

gijsverdoeskleijn:
Gijs-to-self: https://gitlab.astro-wise.org/micado/micadowise/-/issues/204

PR #86, opened 2022-10-06
Kl/gijs notebook

astronomyk:
@wjvriend @hugobuddel @oczoske 

Changed the incremental header keyword symbol from `§` to `++` to avoid header bugs in MicadoWise resulting from scopesim output. 

See accompanying scopesim PR: https://github.com/AstarVienna/ScopeSim/pull/166


hugobuddel:
How does the § break MicadoWISE? The § should never actually enter the header keywords as it should be replaced by the detector number. This does require a correct combination of ScopeSIM + the IRDB to be used together, because otherwise the replacement is not done.

It seems someone is either using an outdated version of the IRDB, or an outdated version of ScopeSIM. So it would be simpler to just instruct people to use the correct versions, because they would still have to upgrade to use the code in this MR, and if they upgrade anyway, then this MR is not necessary.

Also, removing the inst_pkgs symlink will break the notebooks, and/or the notebook tests.

MICADO/docs/example_notebooks/MICADO_4_stars.ipynb
```diff
@@ -0,0 +1,256 @@
...
+   "id": "e962001c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Gijs's local computer\n",
+    "# sim.download_packages([\"Armazones\", \"ELT\", \"MAORY\", \"MICADO\"])\n",
+    "\n",
+    "# Kieran's computer -> redirects ScopeSim to folder containing IRDB \n",
+    "# MicadoWise could do this too and clone daily from github to save running time\n",
+    "sim.rc.__config__[\"!SIM.file.local_packages_path\"] = \"F:\\Work\\irdb\""
```
hugobuddel, line 39: The inst_pkgs symlink made it not necessary to specify a path, because it would just use the local checkout, so you would always run the notebook with the version of the irdb that the notebook was in, thereby preventing having paths to peoples local computer in the notebook.

Not sure how this works on Windows, but windows supports symlinks as well, so I guess it should work.

MICADO/docs/example_notebooks/MICADO_4_stars.ipynb
```diff
@@ -0,0 +1,256 @@
...
+   "execution_count": 3,
+   "id": "418bfe52",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "cmds = sim.UserCommands(use_instrument=\"MICADO\", set_modes=[\"SCAO\", \"IMG_1.5mas\"])\n",
+    "\n",
+    "# EXPTIME = 3600 = ndit * dit\n",
+    "cmds[\"!DET.dit\"] = 30\n",
+    "cmds[\"!DET.ndit\"] = 120\n",
```
hugobuddel, line 61: For a proper observation with correct headers, it is required to put some more information as input to ScopeSIM, see the other notebooks as examples.

MICADO/docs/example_notebooks/MICADO_4_stars.ipynb
```diff
@@ -0,0 +1,256 @@
...
...60b58BQ+l+ZAgIIFaTJlH51RwHXL4AuYAWqJTefxYVb2QccsBzSz6yc114gAEcm1ezw7CC15tSwLlBFY3Gm5ZS6PCqAFhDleogZHbgJbCc6za5XdN8gdrxGa8dHzfedRXYh5o8gyUEtAFlCqetiAf1m2taf7RS81idBJb1WLbsBquBO7WmBDkCVHeK4OXswUBaFBH+uY1jfn0CpGEzu2foeAb3xwoJLeURrDSiTdgN09W4Z7K+t3SfsMT1emYXSWjh0sGKmV883FoDKJghmtbfE9n4UeH7abnXLQDWzeGU7b36frOPWGoFymcHtO431frROCGT5u5ZpAA7AxstogK+5GBWEGzzPkvDZKtHSvOj22SJm7veWBPDeXAAq1pjvIBVscg6M582/h4QZVktcqFlWMO8AnkHIfHt1jUu8O9zIdlwnwrPMrZCVabmOliwCS+Yo2q5vzWZ9vj6Z66MHvNq4Xdyg9SdQwQMEgDfhuJRgPe9uDxZvSGPZWALFhnEhylfSlcjcLCu9OHsz0obCiHcqfXURyEqyVDDYsnh+uTezAlH6LkswtNl+ENEqMp/sUiOAGuewTg8GubkU2oWLMYKpvzW1X1XfE0oTinHtiDkbHBOgTW3ycluw7ADY3+zUMKM3nh3lg/ga/NHSvMXOQIMgX2BtsBhk7cMeydamAUBdoPRux6JtXkCFtHT1rmYBxedDvSf+vmx2as5hYV4bYOWBaU4Z5fM9V7epJfXGAq15P9tB4GXrQqA9/HcuBNmZctaBqd/Tga7AQ9Xrk4VgZkzrj1nTANlAHq4Eh9anZs/JeSpLKttdEfTIo1kA1jIXBj66d9l8JQFE3nV63W4Ks975fN34TkYfOlxX+bDfFj+tNnFcj64CF8jiWAE8kGVNJzcVXcLfuwLNJQAYXKdvrwRWtwdiqmiN8hpLOI7Z3fbe7Pap9XpAs19VqAfPRxclUYetOZxBNpuZmO153Lq9ZM2FKiKatPPi2rxEAQUVZQ2UZ+oenbVYx4T7ZT7Q0kltf51JqHHM+726TpKB7XnOObBIWso+sCQbBk6ra4lAj5h47SGqxkcHBLJJsxXVWPYSoLMkqKDN8+9+X4qepoPXJiCm7x8B1xbceJTrEoiiL7Dlanu5XUzt6d7nyznBkpYEtneLq4X2SYBthbOD40sDPxw7k9ZnQRTrybkC4Lawk6yqvyPhD
```
hugobuddel, line 212: This should all just work, but apparently there are other things broken and therefore none of the notebooks work, see https://github.com/AstarVienna/irdb/actions/runs/3150371248/jobs/5123073223 .

So it is effectively impossible to review this PR, because of the ScopeSIM (or ScopeSIM_Templates) bugs that need to be fixed first.

My suggestion is to first ensure that all the notebooks run on the dev_master branch, and only then try to add new notebooks.

MICADO/test_micado/test_micado_fits_headers.py
```diff
@@ -0,0 +1,16 @@
+from astropy.io import fits
+import scopesim as sim
+
+
+class TestHeaders:
+    def test_extension_incremental_headers_count_up(self):
+        sim.rc.__config__["!SIM.file.local_packages_path"] = "F:\Work\irdb"
```
hugobuddel, line 7: Fixed paths that refer to your local computer should not be in tests.

hugobuddel:
The tests fail because the "++" is not converted to a number:
```
        for i, hdu in enumerate(hdul):
            if isinstance(hdu, fits.ImageHDU):
>               assert hdu.header["EXTNAME"] == f"DET{i}.DATA"
E               AssertionError: assert 'DET++.DATA' == 'DET1.DATA'
E                 - DET1.DATA
E                 ?    ^
E                 + DET++.DATA
E                 ?    ^^
```
This is expected, because the § was not converted to a number before, so there is no reason to assume that the ++ would be.

I'm fine with reviewing PR's, but please first make sure that all tests pass before assigning a PR, unless specific feedback is requested about a failing test.

hugobuddel (changes requested):
I've added some inline comments to the notebook as well. It might that these are collapsed in the diff view because the file is too large, please press 'load diff' to see them.

MICADO/test_micado/test_micado_fits_headers.py
astronomyk: Done! Thanks for picking that up

PR #89, opened 2023-02-15
Fix OSIRIS MAAT tests

hugobuddel:


hugobuddel:
Merging, because 0d36bbf2bdbe38c705849e4d25f0acc35d6b4e5e and db3feef2177a485f25b13a0559487534ee69da93 should not have been merged as-is, because they broke the CI

PR #90, opened 2023-02-16
Fix MICADO notebooks

hugobuddel:
The notebooks should now work again.

The CI will now force `USE_FULL_DETECTOR = True` in the notebooks before running them (through sed). This will allow us to have full notebooks as examples, but still allow them to be ran on the CI.


hugobuddel:
ScopeSIM_Templates 0.4.4 and  ScopeSIM 0.5.5 have just been released, so this should now work. Merging to see whether the CI works on dev_master

PR #95, opened 2023-03-21
Oc/update metis demo notebooks

oczoske:
- In all demo notebooks `download_package` has been replaced with `download_packages`. The command is actually commented out so it does not make any difference in automatic testing.
- In `demo_slit_wheel` and `demo_filter_wheel` the new methods `add_slit` and `add_filter` are now demonstrated.
- `demo_metis_lms_efficiency` has been updated and should now be able to find `TRACE_LMS.fits` by itself...

A comment: The notebooks seem to be set up so that automatic tests work. However, the main purpose of these notebooks is to be used interactively by users so as to instruct them on specific effects. In order to do so they need to comment out the local path command and uncomment the `download_packages` command. While the effort is small, it is still somewhat awkward.

### DIFF
--- a/METIS/docs/example_notebooks/demos/demo_adc_wheel.ipynb
+++ b/METIS/docs/example_notebooks/demos/demo_adc_wheel.ipynb
@@ -46,7 +46,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# sim.download_package([\"instruments/METIS\", \"telescopes/ELT\", \"locations/Armazones\"]) "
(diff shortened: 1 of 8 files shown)


hugobuddel:
Thank you @oczoske . 

It seems I did not yet merge #94 , to run the METIS notebooks in the CI. (I couldn't check yesterday.) There is a small bit of overlap with this PR (related to finding the files). I'll merge yours first and then fix #94.

The CI fails, but that seems to be due too a problem with OSIRIS, so that should not stop this PR. I'm rerunning the tests on the dev_master branch, to see whether the problem occurs there as well, afterwards I'll merge this.

PR #97, opened 2023-03-21
Improve CI run

hugobuddel:


teutoburg:
I see this just adds the markdown link checker. Why not just call the (almost identical) workflow from our DevOps repo? Actually it seems in many places we're not actually using the DevOps one, but just a copy-pasted version thereof...

hugobuddel:
Using the one in DevOps doesn't really save anything. It would just change the last line to point to the DevOps one, which then would point to the `gaurav-nelson` one. So there isn't really a point to having the `workflow_call` trigger in the DevOps workflow.

hugobuddel:
As in, I was originally hoping it would be possible to somehow 'import' the whole yaml file from the DevOps repo, as you can do on gitlab, but apparently that is not possible with github.

teutoburg:
I see, doesn't really make a difference. Anyway, shall we merge this in?

teutoburg:
Oh, just noticed this is still a draft. Did you have more in mind for this PR?

hugobuddel:
I probably planned to add more, but I forgot what that would be. Thanks for cleaning this up

PR #100, opened 2023-05-23
added tables from ERIS user manual

masha-astro:
Tables from this user manual: https://www.eso.org/sci/facilities/paranal/instruments/eris/doc/ERIS_User_Manual_v112.0.pdf
I included only tables with some numbers, not all tables from this pdf.
Is it the right format for these tables?


hugobuddel:
Great start!

Most tables 'should' have some kind of header. Some values in the headers are actually used, e.g. units are often defined in the header. Just look around for similar tables. For example most filters have a header like:
```
# author : Ric Davies
# source : Ric Davies
# date_created : 2017-11-20
# date_modified : 2017-11-20
# status : Design, pre PDR list of filters
# type : filter:transmission
# center : 1.6351241752644228
# width : 0.28732440255992575
# blue_cutoff : 1.49146197398446
# red_cutoff : 1.7787863765443857
#  
wavelength transmission
1.432 0.0000
1.433 0.0000
1.434 0.000100186
1.435 0.000100527
....
```

All these tables are to be used in some 'Effect' in ScopeSim. This association between the ScopeSim effects and the data in the IRDB is done through yaml files that are also part of the IRDB package. The two yaml files you should start with are:

* `default.yaml`, which describes the whole thing, e.g.
  * what other IRDB package you'd need (probably `VLT` and `Paranal`)
  * what other yaml files need to be imported (probably `ELT.yaml`, `Paranal.yaml`, `ERIS.yaml`, '`ERIS_detector.yaml` (or so))
  * default properties of an observation that one often wants to vary
  * which modes the instrument has, maybe just start with a single mode
* `ERIS.yaml`, which contains the description of (only) the instrument, for example
  * which filters it has, and what filename convention is used for the filters, thereby linking to your filter files
  *  which mirrors it has, described in some other file that you might not yet have
  * which other effects belong to the instrument, and which files are necessary for those
* `ERIS_detector_something.yaml`, one file for each detector array, stating e.g.
  * the geometry of the detector,
  * the quantum efficiency,
  * the nonlinearity,
  * etc.
* mode-specific yaml files, which include effects that only apply to that specific operation mode


So all of these tables have to be referenced in these yaml files somehow. Let me go through the list and make a guess where the table should probably be used:

* `nix-filters/*`: through `ERIS.yaml`. Perhaps you could prefix a `TC_` to the names, from 'transmission curve' and perhaps remove the `filter_` prefix.
* `FOV_and_spaxel_size_for_each_pixel_scale_options_of_SPIFIER.dat`: I think this information should go into the different mode yaml files
* `master_bpm_lamp.fits`: This seems like a bad pixel map, and there is currently no Effect in ScopeSim that can add bad pixels in this way. But such an effect would be usefull. 
* `Maximum_star_brightness_permitted_for_each_filter_camera_and_readout_mode.dat`: I don't think this is needed to run the simulator, it seems more like a check when preparing the observation. Maybe the information can somehow be converted into input for the `LinearityCurve` effect so the detector gets saturated at the right moment.
* `Maximum_star_brightness_permitted_for_each_grating_and_plate_scale_configuration.dat`: similar
* `NIX_filter_wheel_contents.dat`: That information should go directly into `ERIS.yaml`, or maybe `ERIS_NIX.yaml` (are the filters specific to each arm?)
* `Positions_and_sizes_of_the_holes_of_the_SAM_23_mask.dat`: I'm not really sure what this is. I'm not sure masks with holes are explicitly supported in ScopeSim. If it is essential, then we can discuss it further, otherwise I'd just leave it out for now.
* `Positions_and_sizes_of_the_holes_of_the_SAM_7_mask.dat`: Idem
* `Positions_and_sizes_of_the_holes_of_the_SAM_9_mask.dat`: Idem
* `Relative_peak_intensities_of_various_components_in_APP_PSFs_with_respect_to_the_nominal_diffraction_limited_PSF.dat`: I don't know what his is
* `Relative_positions_and_strengths_of_ghosts_at_the_centre_of_field_with_the_13mas_JHK_camera_in_NIX.dat`: I suppose this would be input to something like a Ghost effect, but as far as I know this is not yet implemented
* `SPIFFIER_grating_parameters.dat`: This is probably  important, but I don't really know how grating is implemented. I can look this up for you if you need it.

The main things that are missing that I can think of now are:
- Geometry of the detector array (size, orientation, etc), you need that to be able to simulate anything.
- Information about the mirrors, like reflectivity and sizes and such. You need at least 1 mirror to make a simulation.


By the way, you should just put the files in the `ERIS` directory directly, not in the `data` directory.

I'm almost out of battery, hopefully this is useful to you.

PR #110, opened 2023-06-30
new MICADO data files

oczoske:
Ric Davies sent new data files for MICADO; these are incorporated into `irdb/MICADO` in this pull request. Comments on changes made to individual files:
-  `TER_entrance_window.dat`: added some meta keywords; added columns `reflection` and `emissivity = 1 - transmission`.
- `TER_coating_antireflection.dat`: this is a new file that is not used by any effect so far. I've added the file and will ask for further information how to use it.
- `TER_SCAO_dichroic.dat`: The dichroic is used in transmission for NIR wavelengths (i.e. for science data). This is different from the current irdb definition. The action of the dichroic in `LIST_MICADO_mirrors_static.dat` has therefore been changed to `transmission`.
- filter curves: The new filter curves have been added as they were provided. The new narrow-band filter curves require a broad-band filter to be set in another filter wheel, otherwise there will be huge continuum leakage from broad wavelength ranges away from the actual filter location. This currently requires the user to be very careful with their setup (which should follow the actual MICADO setup closely).  A proposal to make this more user friendly is deferred to  #112.


teutoburg (approved):

hugobuddel (approved):
Thanks @oczoske.

About the curves: most of the original curves are such that they drop to 0.0 at the edges, and the new ones don't, some of the broad ones even go back up to 0.1 at the edges. I don't think this is technically problematic, so if you think this is fine, then I'm okay with it.

I'll let you decide whether you want to change the edges of the curves or not. Feel free to merge.


A tangential thought about tests: it would be good if we somehow start adding some tests that validate (and demonstrate to users) that ScopeSim actually performs according to Ric's specification. (Or not, in case of the antireflection coating.)

For the filters perhaps something like simulating an LSS observation of a featureless star, and then verify that the center and FWHM of the spectrum matches that of the filter. But that is beyond the scope of this PR.


The best would be if Ric (or Joost, etc.) accompanies these new tables with a (quantitative) way to demonstrate that everything works as anticipated, since they know best. We could then translate that to Python tests.

oczoske:
- about the curves not dropping to zero: That actually was a problem at some point when extrapolation was done by a constant function at the level of the last point. I _think_ this was subsequently replaced by extrapolation using zero, but I'd like to verify that in the code. As the data were provided by Ric I would like to touch the data themselves as little as possible.
- There seem to be some performance tests. At some point a test on limiting magnitudes failed (due to a clerical mistake - that missing underscore). I was actually a bit surprised that this test passes with new transmission data...

hugobuddel:
* I agree that it would be good to keep our files as close to Ric's versions as we can. So perhaps it would indeed be better to extrapolate by using 0. I don't know what the state about this is.
* A while ago I relaxed the criteria for some of those tests, or marked them as `xfail`  or `skip`, because I wanted no red tests. Maybe we can tighten them again. But if they would/should indeed fail with these new curves, then we should mark the tests as such (`xfail`).

oczoske:
Looking at the filter curves again, the only suspicious one is K-long, and even that is to be used in conjunction with K in the other filter wheel. Therefore, in the current setup everything is effectively forced to zero if used as described in the operational concept.

hugobuddel:
Those background tests are probably the main ones. I didn't really know what the correct numbers should be.

There is also this limiting magnitude test https://github.com/AstarVienna/irdb/blob/bed2a8c865bbdbb09627b654e72882fa4a3a3037/MICADO/test_micado/test_micado_imaging.py#L53-L58

where I changed the limit for the J-band.

Shall we merge this?

Next Tuesday me and @teutoburg are going to make some new releases of the ScopeSim-related software packages, and then maybe also of the IRDB packages. Do we want this in before that?

oczoske:
I've just gone ahead and merged it. Good to have new releases out - this includes all of the MICADO requests for spectroscopy for now.

PR #115, opened 2023-07-11
7-Eleven release

teutoburg:
Update version files of:

- Armazones
- ELT
- METIS
- MICADO
- MORFEO

Pushed those versions to the InstPkgs server.

This should also include the corresponding tags, but since I'm pushing this from my fork, those will probably not be included. Worst case I'll add them manually afterwards...

### DIFF
--- a/Armazones/version.yaml
+++ b/Armazones/version.yaml
@@ -1,3 +1,3 @@
 release: stable
-timestamp: '2022-04-26 17:04:12'
-version: '2022-04-26'
+timestamp: '2023-07-11 10:03:42'
+version: '2023-07-11'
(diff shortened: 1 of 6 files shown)


teutoburg:
Tests failing because there's apparently a test (that I missed) checking the version numbers in the now deprecated yaml file. This should obviously be sorted...

hugobuddel:
Thanks, merged! I'll merge it into master as well, and I propose that we then tag the master branch (I'll do that), even though that is not entirely correct.

This highlights another problem that I forgot to address yesterday. I find it pretty annoying that releasing a new version updates the version files, because that makes it effectively impossible to release from a specific tag. Therefore I added the `-d` flag to the publish script to make it possible to read from the version file instead.

I believe we are also reaching consensus that we don't really need the `dev_master` branch as an inbetween step. My proposal for the release process in the future is:
1. Merge everything we want to release into `master` 
2. Update the versions in the yaml files
3. Tag
4. Publish

PR #117, opened 2023-07-13
make FDR slits available

oczoske:
- new slit mask files added for the FDR slits 3000x16, 3000x48, 15000x20 according to ELT-TRE-MCD-56300-0014_3.0
- the new slits are made available through new instrument modes defined in `default.yaml`. The old ones are still there, with a newly added deprecation notice. It would be nice if `scopesim.UserCommands` would display these.


hugobuddel:
Good! I can't really judge the contents of the files, so I trust you know what you're doing.

Going from "source : My imagination" to "source : ELT-TRE-MCD-56300-0014_3.0" is definitely a good thing. (Which means I could probably verify the contents of the files..)

I'll merge and create a release.

PR #127, opened 2023-08-01
Better report badges

teutoburg:
This is a (rather large) rework of how the package report badges are created and manipulated.

## What's changed

In no particular order:

- The report badges are now handled by a context manager (`badges.BadgeReport`) instead of the individual functions previously used. The details of how to use it are given in the class's docstring.
- The individual badges are now different subclasses of a common bas class `badges.Badge`. The dispatch of which subclass is used happens in the base class, based on the value pass to it. The subclasses are not meant to be created manually, although it should not cause any issues to do so. Again, details in the docstrings.
- For simple messages, a message-only badge is now used, e.g. [](). This is considered a bit clearer than the previous e.g. [](), especially in cases where the "opposite" was never produced (e.g. []()), because the badge denotes the absence of some other badge(s).
- Some tests produced a positive badge even if no file was found in that package to be checked (and thus no check *could* fail in the first place). This was a bit misleading, so at least some of these tests now produce a more "neutral" badge informing stating that this was the case.
- For badges in nested dicts, a sub-heading is now produced up to a nesting level of 2, anything deeper (currently not used) will still produce a nested list. Looks a bit nicer in my opinion, but that might be personal preference...
- Some (rather empty-ish) packages were previously not recognized by `utils.get_packages()`, because that function relied on a `.yaml` file being present in a package with the same name as the package itself (referred to as a "self-named yaml"). This created the weird situation that one of the tests checked if this self-named yaml was present, but only got packages as an input where that was the case. This was noted in the test though. That situation is now improved: `utils.get_packages()` no longer relies on the presence of a self-named yaml, but rather will recognize any directory in the repo as a package, that: a) does not start with a `.` or `_` and b) is not one of either `irdb` or `docs`. These special folders are currently hardcoded. There may be a better way to do this, but I think for now it's an improvement over the previous situation.
- Any error or conflict produced in the `.dat` files also logs a corresponding error message. These logs are captured and stored in an additional file.
- The markdown badge report now has a header which includes the time and date of creation.
- The globbing used to find all the `.dat` files has been improved. Previously it did not find any such files in subfolders of packages. This results in a bunch of now failing files.
- Many of the tests are now better parametrized. Previously, they looped over all packages, collecting any errors, and finally asserting that the list of errors is empty. Now, *each package is a separate test* using parametrization. This not only simplified the structure of the tests themselves (some were nested up to 8 levels), but also means it's much more obvious in pytest itself, how many things pass and fail.
- All parametrized tests where at least one of them currently fails are marked with `xfail`, to not pollute the entire test suite with rather trivial issues in some data files.

## What's new

All `.dat` files in all packages are checked for consistency of the dates given in the file header, specifically `date_created` and `date_modified`, as well as the dates listed under `changes`. The latest date listed in `changes` must match the `date_modified` (which must then exist); if there is a `date_modified`, at least one change must be listed with that date. Finally, `date_modified` must be a later date (or equal to) `date_created`, if `date_modified` exists. If any of these conditions is not met, this file is listed in the report with a `conflict` marker.

### Useful and not too hard

- More general information of the packages, like version numbers, could be integrated in the badges.
- The badges for each package could also become part of that package's `readme`, to keep it all in one place.
- Currently, the report must be updated and pushed manually. This somewhat defeats the purpose. Ideally it would be created as part of the CI/CD. More on that later.

### Nice-to-have, more complicated

- Some tests depend on each other, in the sense that if one fails the other will not be able to test what it's supposed to, and should be skipped. E.g. if a `.dat` file produces an error upon loading, there's not much point in checking its modification dates. This can be done in pytest, but needs some more work, and might be difficult with the parametrization.
- Some tests, despite being parametrized with respect to the packages, still "loop and collect" to finally assert an empty error list. This could be changed to multi-stage parametrization, but that might make it harder (though most likely not impossible) to collect the "all good" style badges.
- The implementation of the tests for the `.dat` files is a bit clumsy and could be improved.
- Several minor TODOs are given in comments in the code.


_REPORTS/badge_report_log.txt
oczoske, line 1: I guess this shouldn't be in the repository. Add to .gitignore.

oczoske (commented):
I don't even know what these badges are and I don't really want to get into it that right now. Hence no detailed review from me.

_REPORTS/badge_report_log.txt
teutoburg: `_REPORTS` is in the `.gitignore`. I specifically "whitelisted" the `badge_report_log.txt` file, because it includes the actual useful error messages associated with the "error", "conflict" etc. badges (and only that). The idea was to separate these from the badges, to keep those tidy. But I'm open for better solutions...

astronomyk (approved):
All looks good to me. I haven't read the code through thoroughly, but if it works as you intended, then I'm happy for this to be merged. This is not "system-critical" code, so I'll take the flak if this merge breaks something down the line.

Thanks for putting so much effort into this particular developer wank ;)

PR #130, opened 2023-08-21
Fix documentation badge pointing to scopesim_templates...

teutoburg:
How did that fly under the radar until now O.o

Anyway, doesn't fix the fact it's still failing...


hugobuddel:
Good catch.

I imported the irdb on my personal readthedocs account (as irdbhb), and indeed the build fails. Or well, so far the job has not yet completed. So I think it is just taking too long. I'll create an issue for that.

Merging this.

hugobuddel:
Created #131

PR #133, opened 2023-08-29
Fix default filter for 3000x16 mode

teutoburg:


teutoburg:
This is trivial and was discussed in the linked issue, so I'll go ahead and merge.

PR #134, opened 2023-09-08
Delete symlinks before building documentation

hugobuddel:
This will delete the `inst_pkgs` symlinks, thereby making RTD work.

However, this changes means that RTD cannot actually run the notebooks anymore. Nevertheless, for most notebooks it would probably not be possible to run them on their servers anyway. And this is better than not having documentation at all.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -30,6 +30,7 @@ sphinx
 
 import sys
 import os
+from pathlib import Path
 
 sys.path.insert(0, os.path.abspath('/'))
 sys.path.insert(0, os.path.abspath('docs'))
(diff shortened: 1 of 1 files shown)


teutoburg (approved):
If it means more things work, which are now broken, I'd say go ahead 👍

hugobuddel:
I think we should build the documentation ourselves on one of our own machines, and then upload those to RTD or something like that. (Which kinda defies the point I suppose.) Because then we can remove the figures from the notebooks in the git repository (e.g. add some pre-commit hook or so) but still have them in the documentation, so we also have a gallery. So this change is just something to get everything working (/not failing) in the interim.

PR #139, opened 2023-10-19
Mostly adaption following the off-by-one fix

teutoburg:
Mostly updates to notebooks, a PSF FITS coordinates fix, and some minor adjustments related to off-by-one.

New versions of MICADO, METIS and WFC3 packages will be created from here.


hugobuddel:
Hmm `master` and `dev_master` of ScopeSim and ScopeSim_Templates now should be identical, so not sure why the job_master tests are failing...

hugobuddel:
Oh those failed jobs were ran quite long ago, so that's why they failed. I'll let the notebooks and one of the job_master tests run and then we can merge it

teutoburg:
The job_master still fails. Merge anyway?

PR #143, opened 2023-11-10
Use more specific lookup reference for default temperature

### LINKED ISSUE #142: In the user command the !ATMO.background and !TEL.temperature and !RO.temperature values are not considered
I tried to change independently the variable affecting the 

1. magnitude of the sky background
2. the temperature of the telescope
3. the temperature of the Relay optics

none of these affect the result.
Only the !ATMO.temperature affects the results. Please note that 2 and 3 may depend on the definition of the system in the IRDB, however, the sky background ( !ATMO.background.value) should be independent. 

Here an example, testing also random noise generation
[Test random seed.zip](https://github.com/AstarVienna/ScopeSim/files/13270469/Test.random.seed.zip)

teutoburg:
Might resolve #142

### DIFF
--- a/ELT/LIST_ELT_combined.tbl
+++ b/ELT/LIST_ELT_combined.tbl
@@ -2,13 +2,14 @@
 # author : Oliver Czoske
 # source : ESO-253082_4
 # date_created : 2021-02-22
-# date_modified : 2021-02-22
+# date_modified : 2023-11-10
 # status : Design
 # outer_unit : m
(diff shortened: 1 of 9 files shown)


hugobuddel (approved):
Seems sensible to redirect the temperature like this. In the end it still all defaults to `!ATMO.temperature` it seems, but now it is at least possible to change the temperatures in a meaningful way.

PR #145, opened 2023-11-20
Run notebook tests in parallel

teutoburg:
- Split notebook job in MICADO and METIS to run them in parallel. Setup is still copy-pasted, because caching takes too much brain to do right, and the execution time is by far dominated by the actual running of notebooks, so it doesn't matter.
- Add Python 3.12 to (non-notebook) test matrix. This actually caused one error, which is now also fixed.
- Add `slow` marker for pytest to the slowest few (non-notebook) tests, which is not currently used by the CI. Also caught one case where the `webtest` marker should have been applied, but wasn't.
- Some small improvements to make the CI runs nicer to look at.
- Prepare to do something smarter with the "Badge Report" files.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: ['3.8', '3.9', '3.10', '3.11']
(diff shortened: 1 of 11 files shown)


hugobuddel (approved):
Great! Thanks for speeding things up

PR #147, opened 2023-11-28
Replace file_name with filename in Effect kwargs

hugobuddel:
`file_name` as a kwarg does seem to work, but it behaves differently from `filename`. In particular, with `file_name` the absolute path of the file is used, and with `filename` the relative path (or well, the actual given path). I have not investigated why.

The reason for harmonizing is that when generating FITS keywords, we'd like to have them to have the filename as it was given by the user or the irdb, not the absolute path where this file happens to be on this particular installation.

Do we perhaps want to force `file_name`?

E.g. put an assert in `data_container.py` instead of a fallback: https://github.com/AstarVienna/ScopeSim/blob/36df1767c2b8143a199d085cdf1af181660c058a/scopesim/effects/data_container.py#L63-L64
```python
        if filename is None and "file_name" in kwargs:
            filename = kwargs["file_name"]
```
to
```python
        assert "file_name" not in kwargs, "Please use filename instead of file_name"
```
or maybe a deprecation notice would be better.

But that is a ScopeSim problem; this PR is just making the IRDB consistent.

### DIFF
--- a/HAWKI/HAWKI_H2RG.yaml
+++ b/HAWKI/HAWKI_H2RG.yaml
@@ -37,7 +37,7 @@ effects :
     description : Quantum efficiency curves for each detector
     class : QuantumEfficiencyCurve
     kwargs :
-        file_name : QE_detector_H2RG.dat
+        filename : QE_detector_H2RG.dat
 
 -   name: exposure_action
     description: Summing up sky signal for all DITs and NDITs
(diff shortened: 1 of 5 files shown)


hugobuddel:
Oh this also makes the `quantum_efficiency` naming consistent in METIS, which should perhaps not belong in this PR but is fine as far as I'm concerned.

PR #150, opened 2023-12-22
Disble SourceDescriptionFitsKeywords for MICADO

hugobuddel:
The SourceDescriptionFitsKeywords still cause problems, see https://github.com/AstarVienna/ScopeSim_Templates/issues/70 https://github.com/AstarVienna/ScopeSim_Templates/issues/69

### DIFF
--- a/MICADO/MICADO_H4RG.yaml
+++ b/MICADO/MICADO_H4RG.yaml
@@ -76,13 +76,15 @@ effects :
         noise_std : 12
         n_channels : 64
 
+# SourceDescriptionFitsKeywords are disabled by default because they still cause problems.
 # ValueError: The header keyword 'SIM SRC0 function_call' with its value is too long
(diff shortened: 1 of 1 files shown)


hugobuddel:
Tests fail because of https://github.com/AstarVienna/ScopeSim/pull/314/files#r1434844238

teutoburg:
What's the status of this @hugobuddel ? I just realized that I never reacted to the review request here. Does this still fail? Should we merge it before making new packages? (even if we now only need a new version for METIS, this still requires a merge to master, so it might make sense to include that as well?)

hugobuddel:
Let's include it. I don't think we fixed this properly. Which is one of the reasons I prefer to use short source names without parameters.

We should keep the ScopeSim Templates issues open probably. And maybe make some minimal examples that show the behavior.

PR #153, opened 2024-02-06
Update of the MICADO detector file for quantum efficiency

### LINKED ISSUE #152: The H4RG quantum efficiency response curve needs an update
The file 
MICADO/MICADO_H4RG.yaml
is currently pointing to a quantum efficiency vs wavelength profile of the old H2RG detector that is not correct (MICADO/QE_detector_H2RG.dat).

MICADO will use H4RG-15 instead.
Unfortunately, the H2RG curve currently used is badly cutting the response in the I-band.
Something like the following (better) matches the expectation:
H4RG
or look here 
https://onlinelibrary.wiley.com/doi/full/10.1002/asna.20230058

carmeloarci:
The MICADO detector yaml now points to the H4RG description and not to the H2RG (previous). The quantum efficiency file is brand new and taken from MICADO FDR data.

/closes #152


hugobuddel:
Thanks @carmeloarci !

After some thoughts (the rest of this comment), I think we should merge this. I'd like to get @astronomyk opinion first though, because I recall Kieran having thought about the MICADO QC curve before.


We now have three different references for the QE curves, and they are not consistent. They are plotted in the figure below. The I-long filter throughput is also plotten, and indeed half of the i-band flux seems to be blocked by the current QE curve.

Different_QE_Curves


1. Apparently our current H2RG curve is from "Finger et al. 2008". I'm not even going to look it up, it is probably outdated. Blue in the figure.

2. You use ELT-TRE-MCD-56309-0004, the "Detector Subsystem Design and Analysis". I have the FDR version of that document, which is 2.0 (2021/02/12), is that indeed the latest? I cannot find a figure that contains the QE Curve in the i-band; only figure 7-4 which contains the cutoff from 2.4 um to 2.6 um. So maybe my version is outdated. Your rendering is in orange in the figure.

3. We use ELT-TRE-MCD-56300-0173 "MICADO Imaging and Spectroscopy Reference" (v1, 2023/11/29). These contain measurements of the QE curve at three wavelengths, for 6 of the detectors. Those are plotted as black dots in the figure.

Some points of concern, in decreasing importance:

* The main point of concern I have with using this new QE curve is that we are kinda committed to following the "MICADO Imaging and Spectroscopy Reference", and it is not directly obvious to me that the orange line fits the black points better than the blue line (except for the i-band). I'm fine with either curve, but I want to stay in the clear with the other people. Could you perhaps verify with JUP or Ric that your curve is indeed the one to use?
* I think we prefer all our transmission curves to drop to 0 at the edge. The new curve goes to 1 at the low wavelength limit. Would it be okay to cut of the QE curve below a wavelength of 0.5 um? Not sure whether that is strictly necessary though, probably not.
* For some reason all the tests are failing, but I expect that this is unrelated to your changes. So we should not worry about the tests, but we should make sure.
* Should we just remove the H2RG curve from the MICADO database?


---
Script to create the figure:
```python
from scopesim.effects.ter_curves import TERCurve, QuantumEfficiencyCurve
from matplotlib import pyplot as plt

fn2 = 'MICADO/QE_detector_H2RG.dat'
fn4 = 'MICADO/QE_detector_H4RG.dat'
qe2 = QuantumEfficiencyCurve(filename=fn2)
qe4 = QuantumEfficiencyCurve(filename=fn4)

fni = "MICADO/filters/TC_filter_I-long.dat"
teri = TERCurve(filename=fni)

QEat800nm = [0.79, 0.87, 0.91, 0.75, 0.76, 0.82]
QEat1230nm = [0.89, 0.88, 0.85, 0.74, 0.74, 0.82]
QEat2000nm = [0.91, 0.88, 0.98, 0.78, 0.81, 0.87]

waves = [0.8,] * 6 + [1.230,] * 6 + [2.0,] * 6
QEs = QEat800nm + QEat1230nm + QEat2000nm

fig, ax = plt.subplots(1, 1, figsize=(12,6))
qe2.plot(axes=ax)
qe4.plot(axes=ax)
ax.scatter(waves, QEs, color="k")
teri.plot(axes=ax)
ax.legend(["H2RG (Finger+ 2008)", "H4RG (ELT-TRE-MCD-56309-0004)", "Measurements (ELT-TRE-MCD-56300-0173)", "I-long"])
ax.set_title("Different QE Curves")
ax.set_xlim(0, 3)

fig.savefig("Different_QE_Curves.png")
```

astronomyk:
@carmeloarci I would also be interested in the reason why the H4RG doesn't go to zero QE in the blue end. Is this a new "feature" of the latest HAWAII chips?

@hugobuddel, I'd say we leave the H2RG file for now, but without a reference in the yaml files, just so that we can still test things using the "old" QE. In a future release, when we're sure that everything is ok with the H4RG curve, we can remove the H2RG one.

hugobuddel:
The regular tests work (did not test the notebooks), so if we are in agreement, then I'll merge this. We can update the curve later if necessary.

carmeloarci:
The QE does not go to zero because of the antireflection coating on it which does not block that part. However,
I found an additional plot which goes beyond (less than) 0.4 and cuts just beyond 0.4 microns.
H4RG
H4RG

hugobuddel:
I'll merge this, as this is clearly an improvement and it doesn't have to perfect. Others will be able to find this PR and the associated issue  The test failures are unrelated to this. It is also indeed not necessary to have the curve go to zero.

Thanks @carmeloarci , much appreciated

/closes #152

PR #155, opened 2024-02-26
Several fixes in preparation of the CainMap

teutoburg:
Mostly harmonize the `set_modes`, all in preparation of the UserCommands -> CainMap change in ScopeSim itself.

There's now one failing test because the call signature of `UserCommands.set_modes()` will change, but hasn't yet. TBH, I don't know why in that example this method is called at all, and not just put the modes in the init call. This seems to be the only place the method is called externally anyway...


MICADO/test_micado/test_micado_compiles.py
```diff
@@ -57,7 +57,7 @@ def test_user_commands_loads_mode_files(self):
 
     def test_user_commands_can_change_modes(self):
         cmd = scopesim.UserCommands(use_instrument="MICADO")
-        cmd.set_modes(["MCAO", "SPEC_3000x50"])
+        cmd.set_modes("MCAO", "SPEC_3000x50")
```
hugobuddel, line 60: Did you change this in the opposite direction? It seems the tests fail on this part
```suggestion
        cmd.set_modes(["MCAO", "SPEC_3000x50"])
```

Armazones/tests/__init__.py
hugobuddel, line 1: I think these `__init__.py` files are there so the tests can import from each other, like functions and such. But it seems nothing imports from the Armazones package, so removing it should not cause any problems.

METIS/docs/example_notebooks/RawHeaders.ipynb
hugobuddel, line 1: This notebook only had text output, and I think it is fine to keep those in the repository, because it doesn't really cause any bloat. Not necessary to revert this though.

hugobuddel (approved):
Good! LGTM

PR #156, opened 2024-02-26
Fix METIS LMS / IFU mode

hugobuddel:
Needs https://github.com/AstarVienna/ScopeSim/pull/376


hugobuddel:
#157 is based of this one, so maybe we can merge that directly. Or perhaps review that one and then merge them in order?

PR #157, opened 2024-02-26
No more UserCommands as rc.currsys

hugobuddel:
Counterpart of https://github.com/AstarVienna/ScopeSim/pull/377


hugobuddel:
All tests now pass using ScopeSim branch of https://github.com/AstarVienna/ScopeSim/pull/377 . Have not yet tested all the notebooks.

hugobuddel:
> All tests now pass using ScopeSim branch of [AstarVienna/ScopeSim#377](https://github.com/AstarVienna/ScopeSim/pull/377) . Have not yet tested all the notebooks.

Did run all the notebooks locally

hugobuddel:
All notebooks and tests pass with both ScopeSim dev_master and https://github.com/AstarVienna/ScopeSim/pull/377 , so I suggest we
* first merge this, then
* release IRDB packages; at least those changed in this branch, then
* merge https://github.com/AstarVienna/ScopeSim/pull/377

teutoburg (approved):
Store `IFU-example.ipynb` with or without output? Otherwise, LGTM.

teutoburg:
> * release IRDB packages; at least those changed in this branch

I can do that part 👍

hugobuddel:
> Store `IFU-example.ipynb` with or without output? Otherwise, LGTM.

It would perhaps be more in line to remove the output. But perhaps it is good to have at least one copy of the output in the git history? Then we can always go back to that version to see what it should approximately look like? It's only 250kb

hugobuddel:
I'll merge this. We can remove the output of the notebook later; which does mean it is in the history forever, but it isn't that large to store one version of it.

I think we only really need a new package of METIS; could you please make that @teutoburg ?

hugobuddel:
So apparently the new IFU example notebook fails with the old ScopeSim. So I probably ran them incorrectly locally and perhaps other METIS notebooks fail too with ScopeSim 0.7.1. But perhaps we can't prevent that. So I'd say that if MICADO runs fine, then it's good enough.

PR #158, opened 2024-02-28
Remove support for Python 3.8

hugobuddel:


teutoburg (approved):
Thanks! You could even now remove the `from typing import List, Tuple, Set, Dict` stuff, because from Python 3.9 onwards, it understands the standard `list, tuple, set, dict` in that context. But the typing import will still work, so it's also fine to do that later 🙂

hugobuddel:
> Thanks! You could even now remove the `from typing import List, Tuple, Set, Dict` stuff, because from Python 3.9 onwards, it understands the standard `list, tuple, set, dict` in that context.

Let's do that later then. I just did an `rg -. "[^0-9]3\.8[^0-9]"` and cleaned up the places where you left comments.

PR #159, opened 2024-02-28
Include symlink for METIS notebooks directory

hugobuddel:
This should make the notebooks in METIS run without having to specify `!SIM.file.local_packages_path`


teutoburg (approved):

hugobuddel:
The new METIS notebook apparently does fail, which I don't understand, "it works on my machine", but that is unrelated to this PR: https://github.com/AstarVienna/irdb/actions/runs/8085838131/job/22094247854

PR #165, opened 2024-03-25
Replace all remaining file_name with filename

teutoburg:
I guess those were missed in #147? Anyway, they caused some errors and also a hard-to-trackdown weirdness of wrong detector layout (imaging layout was used in IFU mode). This fixed it.

### DIFF
--- a/METIS/METIS_DET_IFU.yaml
+++ b/METIS/METIS_DET_IFU.yaml
@@ -4,9 +4,10 @@ object: detector
 alias: DET
 name: metis_lms_detector_array
 description: A set of 4 H2RG detectors
-date_modified: 2021-12-16
+date_modified: 2024-03-25
 changes:
   - 2021-12-16 (OC) some rearrangements
+  - 2024-03-25 (FH) filename instead of file_name
 
 properties:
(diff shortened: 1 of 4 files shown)


hugobuddel (commented):
These `file_name`s should not cause any problems. In fact, we changed some of these before (not sure whether in #147 or later), but we had to revert them (to get the IFU mode working).

The `__init__()` of effects should not have an `file_name` parameter, and therefor the `kwargs` in the yaml files should never have an `file_name` keyword them.

However, these `file_name`s only appear in the `properties`, and thus only in the values of the `kwargs` in the yaml files, not the keys.

Now arguably, we can also change these to `filename`, and I would be fine with that. However, we should first make sure that we understand what we are doing, because these are self-consistent and should not have caused any "hard-to-trackdown weirdness".

In particular, the current ScopeSim code will be broken if we make this change now:
```python
scopesim/effects/metis_lms_trace_list.py
310:        filename_det_layout = from_currsys("!DET.layout.file_name", cmds=self.cmds)
```

Could you elaborate on why this change is necessary?

PR #166, opened 2024-04-03
Add WCU mode to METIS

teutoburg:
The corresponding source objects will be added in `ScopeSim_Templates`.

I ended up splitting the "upstream stuff" (i.e. location and telescope) into a separate mode, for now called "light". I'm open to better name suggestions, the reason I choose "light" was because I know the term "light frame" for anything that observes "through the telescope" (like actual observations, but also things like dome flats) as opposed to "dark frame" without the telescope in the optical path.

The reason for splitting this in the first place (and not just adding the two yamls to all other modes like I did in my first attempt) as that the WCU mode will need to work for multiple different instrument modes (image, lss, ifu, ...) and so the other option would have been separate modes "wcu_img_n", "wcu_img_lm", "wcu_lss_n", etc etc etc. The split is also similar in concept to the SCAO/MCAO + "actual mode" split in MICADO.

I realize that might break some existing code which doesn't explicitly set the new "ligth" mode yet (although the default does). For the time being, a temporary "solution" might be to keep this change in a "dev" version of the METIS IRDB package (that's fully supported after all), so the default `sim.download_packages` will still have the old mode...

### DIFF
--- a/METIS/METIS_WCU.yaml
+++ b/METIS/METIS_WCU.yaml
@@ -1,0 +1,26 @@
+---
+### METIS WCU MODE
+
+object: instrument
+alias: INST
+name: wcu
+description: warm calibration unit mode
+date_modified: 2024-04-06
+changes:
+  - 2024-04-06 (FH) created file
+
+properties:
+
+effects: []
+
+---
+### default simulation parameters needed for a METIS simulation
+object: simulation
+alias: SIM
+name: METIS_simulation_parameters
+description: RC simulation parameters which need to change for a METIS run
+
+properties:
+  spectral:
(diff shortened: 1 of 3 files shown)


hugobuddel:
Good work!

(Next commit: "refactor the different modes into separate yaml files". Commit after that "rename the yaml files to match names of observing templates". :raccoon:. j/k, we should just YAGNI al that, and right now we need this new mode, let's see where it ends up.)

teutoburg:
Tests are failing because the new WCU mode produces an error when the FITS header effects try to access an (obviously now nonexistent) `telescope_reflection` property. This is a simply fix in ScopSim which I'll prepare and then link to this PR...

teutoburg:
Superseded by #212

Issue #167, opened 2024-04-08
METIS LSS scripts for Wolfgang


astronomyk:
To help Wolfgang we should create one (or six) python script(s) that wolfgang can run from the command line that produce the fits files needed by him to investigate the LSF for various observation modes:

Tasks to do:

- [x] (@teutoburg) script(s) to simulate `image_plane` images for LSS with:
	- all effects turned off except atmospheric TERCurve
	- input Source object is a flat spectrum from 3-15µm, 
	- for two spatial objects:
		- point source
		- slit filling extended source (i.e. large image)
	- for filters: L_spec, M_spec, N_spec
	- deliver the image_plane image and ignore the readout crap
	- save the image_plane image to disk

- [x] (@astronomyk) Updated trace geometries for the LSS modes

teutoburg:
Ad 1) Shall the scripts be pushed anywhere or just shared locally?

PR #168, opened 2024-04-08
Kl/update lss traces

astronomyk:
As part of to #167, here are the new FITS files for the LSS traces based on the optical model from 2023-06-20 that was used for manufacturing the LSS grisms.


teutoburg:
I'm not sure I understand the commit history here. Why does this go back 2 years? I checked out a local version of this branch and it let me rebase onto `dev_master` without any conflicts. Is there a specific reason why the commit history is what it is here? If not, I can push the rebased version to this PR if you want.

MICADO/test_micado/test_micado_spec.py
```diff
@@ -67,12 +67,23 @@ def test_runs_spec_hk_15000x50(self):
...
+        # cmds.cmds["!SIM.spectral."]
 
         micado = sim.OpticalTrain(cmds)
         FULL_DETECTOR = True
         micado["detector_window"].include = not FULL_DETECTOR
         micado["full_detector_array"].include = FULL_DETECTOR
         micado.observe(src)
+        hdul = micado.readout()
+
+        if not PLOTS:
```
teutoburg, line 79: Shouldn't that be `if PLOTS`?

WFC3/test_wfc3/test_full_package_wfc3_ir.py
```diff
@@ -171,17 +171,17 @@ def test_actually_produces_stars(self):
         hdu_av = np.average(hdu[1].data)
         exptime = cmd["!OBS.ndit"] * cmd["!OBS.dit"]
 
-        assert hdu_av == approx(implane_av * exptime, rel=0.01)
-
-        if PLOTS:
+        if not PLOTS:
```
teutoburg, line 174: Same as above...

hugobuddel:
This broke the METIS notebooks: 
* https://github.com/AstarVienna/irdb/actions/runs/8615086139
* https://github.com/AstarVienna/ScopeSim_Data/actions/runs/8626248477/job/23644148710 

Please run the notebooks before you merge something big like this @astronomyk

hugobuddel:
The notebooks also run when you start the CI on github manually, so there is no reason to not run them.

hugobuddel:
Also note that while `TRACE_LSS_L.fits` is replaced, with the original being renamed to `TRACE_LSS_L_old.fits`, it was `TRACE_LSS_L_sci.fits` that was used as default trace file. Now the default is the new `TRACE_LSS_L.fits`. I think this is confusing. Is the `TRACE_LSS_L_sci.fits` also old? I think old files should be removed; they can be found in the history if necessary.

Comparing `TRACE_LSS_L_sci.fits` with the new `TRACE_LSS_L.fits`, it can be seen that the main difference is that these header keywords only appear in `TRACE_LSS_L_sci.fits`: {'CTYPE1', 'CRVAL1', 'CRPIX2', 'MJDREF', 'CRPIX1', 'CDELT2', 'WCSAXES', 'CRVAL2', 'LATPOLE', 'CUNIT2', 'CTYPE2', 'CDELT1', 'CUNIT1'}. The other differences in the headers is expected, like dates and such.

`TRACE_LSS_L_sci.fits` is still in the repository, so we could 'fix' the notebook by just explicitly referencing that in `!OBS.trace_file`, but that seems kinda wrong, since I expect that file to be outdated too.

It is also concerning that the `TRACE_LSS_L_sci.fits` a the old `TRACE_LSS_L.fits` could not be used interchangeably.

I don't know where these headers came from. I suppose the best solution would be to add similar headers to the new trace maps. Maybe we can just copy them over? I'm assuming here that the overal structure of the FITS file has not changed.

hugobuddel:
The notebook says

> Before showing the background subtracted spectra, we convert the pixel numbers to wavelength and spatial position along the slit. This is possible because the current version of Scopesim/METIS by default uses perfectly linear mapping of the spectra onto the detector. The WCS keywords to use are in metis['spectral_traces'].meta. 

I have not inspected the files, so I don't know whether the new files also use a "perfectly linear mapping". It seems so, but maybe the numbers are not the same anymore. The traces seem more compact.

teutoburg:
> I think old files should be removed; they can be found in the history if necessary.

agree 👍

PR #170, opened 2024-04-15
Add Quantization effect to METIS detector yamls

teutoburg:


hugobuddel (commented):
Looks good! But we can't test whether it works, because the CI fails due to #168. Maybe we should fix that first, because it would not surprise me if the quantization will also break some notebooks.

teutoburg:
At least one METIS notebook fails with this. I'll get the `fh/rtd` branch merged first, because that one is modifying those anyway...

teutoburg:
See: https://github.com/AstarVienna/irdb/actions/runs/8802083107/job/24160519457

hugobuddel:
We want to release the METIS Simulations tomorrow (Wednesday), so I'll fix this and merge it.

hugobuddel:
casting to uint currently overflows to MAXINT, that is not supposed to happen. Well, negative fluxes shouldn't happen, but if they happen, they shouldn't be cast to extremely high values

hugobuddel:
Fixed the overflow in https://github.com/AstarVienna/ScopeSim/pull/414

PR #172, opened 2024-04-18
Fix workflows after ScopeSim main change

teutoburg:


teutoburg:
I'm merging this to get #174 to pass...

PR #176, opened 2024-04-24
Rename trace file extensions to match grating efficiency filenames

Isa29bella:
This used to be identical in the previously default *_sci.fits trace files, and causes a silent error when attemting to include the trace efficiencies.

Created in cooperation with @teutoburg


teutoburg (approved):

hugobuddel:
And thanks @Isa29bella ! Such a PR is much appreciated

teutoburg:
> @teutoburg should we add a test to prevent such problems in the future when we get new traces?

Absolutely. The issue here (or not really an issue but just something to be aware of) is that the `SpectralTraceList` assumes that the names of the efficiency extensions are identical to those of the trace definitions. See also the docstring of `SpectralEfficiency`.

So actually, all IRDB packages that use traces and efficiencies should have a standardized test that simply checks if those match.

PR #181, opened 2024-07-16
Test whether exptime is honored.

hugobuddel:
See https://github.com/AstarVienna/ScopeSim/issues/438

And https://github.com/AstarVienna/ScopeSim/issues/442 and https://github.com/AstarVienna/ScopeSim/issues/439


teutoburg:
Rebased to include #182, which should fix the failing tests previously seen here.

teutoburg (approved):
Seems like a useful addition 👍

Issue #173, opened 2024-04-19
[METIS] Include rectification in example notebook


teutoburg:
The error occurs because unlike in `TRACE_LSS_L_sci.fits`, the new `TRACE_LSS_L.fits` does not define WCS keywords in the (primary) header. But maybe that's fine, the larger issue behind this seems to be that the example notebook `LSS-YSO_model_simulation` mentions that:

> Before showing the background subtracted spectra, we convert the pixel numbers to wavelength and spatial position along the slit. This is possible because the current version of Scopesim/METIS by default uses perfectly linear mapping of the spectra onto the detector. The WCS keywords to use are in `metis['spectral_traces'].meta`. The plots will be restricted to the area covered by the spectra.

Which is not true anymore. So perhaps we need to include the rectification in that notebook. For now, I'm disabling the part of the notebook that causes problems, so our CI can pass again...

PR #183, opened 2024-07-21
add rectification, needs cleaning

oczoske:
This deals with  Issue #173. 

METIS now uses by default the full non-linear spectral mappings for LSS simulations. The notebook `LSS-YSO_model_simulation.ipynb` relied on the use of the linear "science" mappings for plotting, this now necessitates an explicit rectification of the simulation results.

Currently, rectification is done as a proof of concept for the first simulation. This needs to be expanded and the notebook cleaned up before merging.

### DIFF
--- a/METIS/docs/example_notebooks/LSS-YSO_model_simulation.ipynb
+++ b/METIS/docs/example_notebooks/LSS-YSO_model_simulation.ipynb
@@ -17,7 +17,7 @@
     "sim.bug_report()\n",
     "\n",
     "# Edit this path if you have a custom install directory, otherwise comment it out. [For ReadTheDocs only]\n",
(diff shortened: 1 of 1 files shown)


oczoske:
The notebook runs now, most of the results reproduce previous results (pdf version dated 2022-03-01). There is no rendered for the last section, which was added in May 2023.  The comparison of the flux-calibrated simulated spectrum with the input spectrum shows large-scale agreement, but uncomfortable oscillatory structure at the blue end.
comparison

oczoske:
The notebook now includes rectification at all stages and should run. The oscillatory residuals mentioned in the previous comment will be addressed in a new issue.

teutoburg (approved):
You know a lot more than I do about the expected results of this notebook. I'm happy if it fully runs again 👍

Issue #161, opened 2024-03-01
MICADO notebook 3_scopesim_SCAO_4mas_fv-psf.ipynb is broken


hugobuddel:
The MICADO notebook [3_scopesim_SCAO_4mas_fv-psf.ipynb](https://github.com/AstarVienna/irdb/blob/dev_master/MICADO/docs/example_notebooks/3_scopesim_SCAO_4mas_fv-psf.ipynb) is broken.

The notebook has a section called

## !!! It appears

But nothing appears! This is the image I get:

![itdoesnotappeartrimmed](https://github.com/AstarVienna/irdb/assets/1306453/4bc1026c-583b-4489-80a1-4476f8dad855)

hugobuddel:
The direct cause of the weird image are these two lines:
```
micado.cmds["!DET.x"] = 10       # [arcsec]
micado.cmds["!DET.y"] = 10
```
 The yellow part on the top right of the image does contain the cluster, but it is all washed out due to the limits.

Removing those two lines, or setting them to 0 does show the proper image.

hugobuddel:
Short code to reproduce:
```python
%matplotlib inline
import datetime
import shutil
import numpy as np
import matplotlib.pyplot as plt
from matplotlib.colors import LogNorm
from astropy import units as u

import scopesim as sim
import scopesim_templates as sim_tp

cluster = sim_tp.cluster(mass=10000, core_radius=0.5, distance=8000)

fig, ax = plt.subplots()
ax.plot(cluster.fields[0]["x"], cluster.fields[0]["y"], ".")
ax.set_aspect('equal', 'datalim')

observation_dict = {
        "!OBS.mjdobs": datetime.datetime(2022, 1, 1, 2, 30, 0)
}

cmd = sim.UserCommands(
    use_instrument="MICADO",
    set_modes=["SCAO", "IMG_4mas"],
)
micado = sim.OpticalTrain(cmd)

micado.cmds["!OBS.dit"] = 600    # [s]
micado.cmds["!DET.x"] = 5       # [arcsec]
micado.cmds["!DET.y"] = 5


micado.observe(cluster)
hdus = micado.readout()

fig, ax = plt.subplots()
im = hdus[0][1].data
ax.imshow(im, origin="lower", norm=LogNorm(vmin=0.9*np.median(im), vmax=1.1*np.median(im)))
# plt.savefig("itdoesnotappear.png")
```

hugobuddel:
[...]

PR #185, opened 2024-08-04
Fix downloads in notebooks

teutoburg:
Following AstarVienna/ScopeSim#446.

For the PSF download in MICADO, we might eventually expand the download facilities in ScopeSim, once `pooch` has proven to be (more) stable. For now, I used pooch's "script mode" to move AstarVienna/ScopeSim#332 forward.

### DIFF
--- a/METIS/docs/example_notebooks/IMG_L_N-examples.ipynb
+++ b/METIS/docs/example_notebooks/IMG_L_N-examples.ipynb
@@ -80,7 +80,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "sim.download_example_data([\"HL_Tau_prep_for_Scopesim.fits\", \"AGN_uc0890_image_l12_i090_p000.fits\"])"
+    "paths = sim.download_example_data(\"HL_Tau_prep_for_Scopesim.fits\", \"AGN_uc0890_image_l12_i090_p000.fits\")"
    ]
   },
   {
@@ -103,7 +103,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "input_hdul = fits.open(\"HL_Tau_prep_for_Scopesim.fits\")"
+    "input_hdul = fits.open(paths[0])"
    ]
   },
   {
@@ -154,7 +154,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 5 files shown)


teutoburg:
I also fixed the naming of various `sim.download_packages(...)` along the way. @oczoske , I only saw afterwards that you did the same for at least some notebooks in `oc/issue_173`, so you might want to rebase your branch. Sorry for the complications. In fact, I think I already did the same fixes in #175, which was never quite completed...

hugobuddel (approved):
Approved, but two comments:

- I don't think it is worthwhile to actually run the MICADO notebook and update the output. The output has not meaningfully changed, so this will add a megabyte or two to the repository for very little benefit. I think it would be better to just change the cells and commit those. Then test them, but discard those changes. (Also the MICADO notebook is broken.. #161.)
- These changes highlight another problem about deprecating the list argument, because now the interface is rather ugly in some cases.

teutoburg:
> * I don't think it is worthwhile to actually run the MICADO notebook and update the output. The output has not meaningfully changed, so this will add a megabyte or two to the repository for very little benefit. I think it would be better to just change the cells and commit those. Then test them, but discard those changes.

Agree. I'm not quite sure why the output would even change at all with the same random seeds. Anyway, I now force-pushed those large but useless "changes" away.

> * These changes highlight another problem about deprecating the list argument, because now the interface is rather ugly in some cases.

Disagree on that one. In all cases except the "LSS-YSO" notebook, the only resulting change was the square brackets, which in my personal opinion actually reads clearer without those, but that's more a matter of taste. Either way I think plus or minus a square bracket is negligible.

In the case of the "LSS-YSO" notebook (which I assume was the one you were referring to anyway), the change resulting from the list -> *args modification was from

```python
sim.download_example_data(list(fitsfiles.values()))
```

to

```python
sim.download_example_data(*fitsfiles.values())
```

Where the original `list(...)` was actually redundant, because the previous implementation of `download_example_data` checked for `Iterable` (as it should), which also works for `dict_values`. So the only "real" change is the addition of the `*`.

The remaining "wrap" of

```python
fitsfiles = dict(zip(fitsfiles, sim.download_example_data(*fitsfiles.values())))
```

is some "iterable magic" to get the locations of the cached files back into the `fitsfiles` dict, which would look exactly the same when still passing a list (as the format of the output, `list[Path]`, didn't change).

hugobuddel:
Good, thank you for fixing yet another one of these!

PR #188, opened 2024-08-18
Further clean up YSO notebook, apply units consistently

teutoburg:
Astropy units were used in this notebook in some parts, but not in others, which led to some inconsistencies. During the flux calibration, the background subtraction was scaled with the exposure time of the standard star, which I believe should not be the case. This also leads to a more realistic scaling factor in the final spectrum comparison (unless I'm missing something, scaling the source *down* to the result makes more sense, as the flux should decrease as it passes through the optical train).

Also made some plots look a little nicer and simplified some of the code.

### DIFF
--- a/METIS/docs/example_notebooks/LSS-YSO_model_simulation.ipynb
+++ b/METIS/docs/example_notebooks/LSS-YSO_model_simulation.ipynb
@@ -74,7 +74,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The input data are cubes of three different models of the same YSO, HD100546. We keep the names of FITS files, the `Source` objects and the results of the Scopesim simulations in dictionaries, indexed by short names for the models."
+    "The input data are cubes of three different models of the same YSO, HD100546. We keep the names of FITS files, the `Source` objects and the results of the ScopeSim simulations in dictionaries, indexed by short names for the models."
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


METIS/docs/example_notebooks/LSS-YSO_model_simulation.ipynb
```diff
@@ -611,8 +619,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "std_result[1].data -= bgresult[1].data / exptime\n",
-    "std_rectified = tracelist.rectify_traces(std_result, -4, 4)"
+    "std_result[1].data -= bgresult[1].data / exptime * std_exptime\n",
```
oczoske, line 622: Since `std_exptime = 1 s`, this shouldn't make a difference, but of course it's cleaner to write down the full Dreisatz.

METIS/docs/example_notebooks/LSS-YSO_model_simulation.ipynb
```diff
@@ -847,12 +853,20 @@
     "plt.figure(figsize=(12, 6))\n",
     "plt.plot(det_wave, result_extract, label=\"flux-calibrated simulated spectrum\")\n",
     "# The scaling for the source spectrum has been estimated by eye\n",
-    "plt.plot(src_wave, src_extract * 1500, label=\"input spectrum\")\n",
-    "plt.xlabel(r\"Wavelength [$\\mu$m]\")\n",
+    "# plt.plot(src_wave, src_extract * 1500, label=\"input spectrum\")\n",
+    "plt.plot(src_wave, src_extract * .42, label=\"input spectrum\")\n",
```
oczoske, line 857: Okay, this is a factor 3600. For some reason I cannot find the change that causes this. It's probably hiding in plain sight.

METIS/docs/example_notebooks/LSS-YSO_model_simulation.ipynb
teutoburg: Indeed I noticed that this is just a multiplication by one. However, like you mention, I think this way it's a lot clearer what's happening physically. Also, with all-consistent unit, we would end up with a dimensional problem otherwise 😅

oczoske (approved):
Those changes are fine - even though I cannot spot what caused the change in scale in the last figure. Overall, I find the notebook more and more confusing, the more often I look at it - it's probably always been that way.  Anyway, we shouldn't waste too much time on this.

METIS/docs/example_notebooks/LSS-YSO_model_simulation.ipynb
teutoburg: `exptime` was previously overwritten with `std_rectified[0].header['EXPTIME']` (which is 1 s), but then used to rescale the rectified results of the 3600 s observation, resulting in a net change by a factor of ~3600. I can't seem to link to the line in the rich diff here, but anyway, that's what was happening. I now used `std_exptime` everywhere it was actually meant to be used, to avoid confusion.

PR #190, opened 2024-09-26
Supply cube as hdul and not just file name

teutoburg:
Required for AstarVienna/ScopeSim#405 to not break the METIS notebooks.


teutoburg:
Notebooks pass: https://github.com/AstarVienna/irdb/actions/runs/11048555544

PR #192, opened 2024-10-31
Set PLOTS to False

teutoburg:
The plot from here popped up during collection time (?) when running pytest locally on the whole IRDB repo. This interrupted pytest until the figure window is closed manually.


teutoburg:
The test using 3.9 fails because ScopeSim main no longer supports that. We should remove 3.9 from the test matrix here. This repo doesn't use the DevOps test, so this wasn't done automatically...

hugobuddel:
Oops my bad. I'll just merge this PR OK

I didn't notice this problem because 'for reasons' my default matplotlib backend is Agg or something so the plots don't show unless I explicitly change the matplotlib backend.

Maybe that would be interesting to add to the test runner: set a non-interactive matplotlib backend and use `sed` to turn all the plots on before running the tests. Because there will probably be failing plots. But it doesn't really matter.

PR #194, opened 2024-11-05
Explicitly pass DIT/NDIT as None

### LINKED ISSUE #193: METIS/LSS_AGN-02_simulation.ipynb is broken
The result using current dev versions now looks like this:

Image

For comparison, since the notebook does not have an output saved in the IRDB, here's a run from an older RTD build where it looks more reasonable: https://irdb.readthedocs.io/en/fh-rtd/METIS/docs/example_notebooks/LSS_AGN-02_simulation.html

Not sure what's causing this or since when it occurs. It was not a result of AstarVienna/ScopeSim#482, it was already the same using `main` before that merge.

I put this here even though it's an IRDB notebook, because I belief the issue to be within ScopeSim and probably also affects other things.

teutoburg:
This should fix #193.


hugobuddel (approved):
"Fix" is a big word, but well, it works I suppose (didn't test it)

PR #195, opened 2024-11-10
Add instrument name in LFOA default.yaml

teutoburg:
One of the example notebooks in ScopeSim started producing `ValueError: !OBS.instrument was not found in rc.__currsys__` for the LFOA. This key is usually found in all other instrument packages. I vaguely remember fixing this, not sure if it was in a local branch on my other machine maybe...

### DIFF
--- a/LFOA/default.yaml
+++ b/LFOA/default.yaml
@@ -12,6 +12,7 @@ yamls :
 - LFOA_SBIG.yaml
 
 properties :
+  instrument: LFOA
   airmass : 1.2
   declination : +30
   hour_angle : 0


teutoburg:
Nevermind, the LFOA package seems to have more issues than this one keyword...

teutoburg:
Also, it does actually run, the error messages are printed but the exception is caught somewhere...

hugobuddel:
> Nevermind, the LFOA package seems to have more issues than this one keyword...

> Also, it does actually run, the error messages are printed but the exception is caught somewhere...

Been there, got the T-shirt!

My conclusion was that we should ensure that all the other instruments that now rely on [write_header()](https://github.com/AstarVienna/ScopeSim/blob/a2d22f28dd74194bddc0832b2de8fe7426e056cb/scopesim/optics/optical_train.py#L486) have a FitsExtraKeywords effect so we can just remove that entire function.

Not that I got around to that either. But the exception is annoying.

teutoburg:
> My conclusion was that we should ensure that all the other instruments that now rely on [write_header()](https://github.com/AstarVienna/ScopeSim/blob/a2d22f28dd74194bddc0832b2de8fe7426e056cb/scopesim/optics/optical_train.py#L486) have a FitsExtraKeywords effect so we can just remove that entire function.

Agree. Not sure if I get around to that anytime soon...

oczoske:
Merge or discard? It's not worth having this PR hanging around. If there are further issues with LFOA, we should start with Issues and create a new PR based on those.

PR #197, opened 2024-11-15
Disable all tests in MICADO_Sci

teutoburg:
These should be evaluated for their usefulness elsewhere and then moved there or deleted completely. This is a temporary solution to stop the ScopeSim_Data runs from failing.


hugobuddel (approved):
Yeah, this at least makes the nightly job not fail..

PR #198, opened 2024-11-16
Use lxml[html_clean] instead of just lxml

teutoburg:
This might be able to fix the most recent ScopeSim_Data fail🤞


teutoburg:
It's hard to test if this will solve the issue. Notebook tests run equally well (or not, but that's unrelated) from both `dev_master` and this branch. I'll just merge and see what happens over in ScopeSim_Data...

PR #199, opened 2024-11-16
Update actions versions

teutoburg:
...to silence warnings from workflow runs.


hugobuddel:
If those workflows can give us a warning, they can give us a PR as well!

hugobuddel:
> If those workflows can give us a warning, they can give us a PR as well!

https://github.com/AstarVienna/DevOps/pull/39

PR #203, opened 2024-11-25

teutoburg:
Some of the instrument packages had changes since the last release. We should have new versions for the Workshop / Conference later this week.


teutoburg:
The failing tests are only because the new version of the artifact uploader fails for duplicate names, which occur because the badge reports are created in every combination. This should be fixed at some point (maybe soon-ish to avoid the fails), but is unrelated to the changes to the instrument packages (or indeed any functional part of anything here), so it's fine to ignore that.

PR #205, opened 2025-01-14
Oc/metis wcu alt1

oczoske:
Configuration files for the METIS WCU mode. This creates new modes `wcu_img_lm`, `wcu_lss_l`, `wcu_lms`, etc.


oczoske:
These tests fail because they are run with the main branch of scopesim, which doesn't know about the WCU. So this PR should stay open until https://github.com/AstarVienna/ScopeSim/pull/494 is merged.

hugobuddel:
A small thought based on https://github.com/AstarVienna/ScopeSim/issues/556 

The `AtmosphericDispersionCorrection` is `METIS.yaml`, which is imported also in the WCU mode. The `AtmosphericDispersionCorrection` is commented out, so it doesn't matter at the moment. But I suppose that it will be a problem when someone uncomments the `AtmosphericDispersionCorrection`.

(Note that I believe that the `AtmosphericDispersionCorrection` does not actually correct anything, but that it introduces the residuals after correction, but I could be wrong.)

Maybe we should move the `AtmosphericDispersionCorrection` to it's own yaml file? Or to the Armazones package?

We don't have to solve that problem now (because it is commented out), so I'm just noting this for future reference. Maybe we should already add a note to  `METIS.yaml`?

hugobuddel:
Ah ok, I thought the idea was to put the AtmosphericDispersionCorrection back in at some point and wanted to ensure that would not break the WCU mode.

teutoburg:
The now failing tests should work after #210, _if_ this branch was renamed to `oc/metis_wcu` (matching the one in ScopeSim), which would necessitate deleting the current `oc/metis_wcu` branch in the IRDB (which as far as I understand is now obsolete anyway?), and then either merging in the current `dev_master` or rebasing onto that. But those are @oczoske's branches, so I won't mess with them (unless asked to do so).

PR #206, opened 2025-01-26
download_packages instead of download_package

hugobuddel:
This is untested; need to run the tests on github


hugobuddel:
Test with notebooks running as https://github.com/AstarVienna/irdb/actions/runs/12977920099

teutoburg (approved):
I'm surprised the METIS and MICADO ones were still there. I might have an unmerged branch somewhere? Whatever, this should be good. Or we'll see if it isn't...

teutoburg:
> So it creates the badge when running the tests against Python 3.10, which goes well, then again for 3.11 and 3.12 and those fail because the badge is already there.

Oh yeah, that needs to be changed. Maybe just reactivate the badge report thing for now, I can come up with a more permanent solution tomorrow.

PR #209, opened 2025-02-15
adapt notebooks to PR scopesim_templates 124

oczoske:
Interface to `flatlamp()` changed in https://github.com/AstarVienna/ScopeSim_Templates/pull/124. If the removal of the wrapper in that PR is rejected, the notebooks should call `micado._flatlamp()` rather than `micado.flatlamp()`.


oczoske:
Discard.

PR #210, opened 2025-02-17
Use branch of same name in ScopeSim and Templates if exists

teutoburg:
This would avoid situations like we have in #205 (well, this wouldn't fix that one in particular, because the branch names don't match, but if they did, it would). The reverse operation (cloning the right IRDB branch into ScopeSim) needs to be done a bit differently, but this is a beginning.

For reference, I took inspiration for this from https://github.com/hse-project/hse/blob/6d5207f88044a3bd9b3539260074395317e276d5/.github/workflows/builds.yaml#L120-L148.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -50,7 +50,7 @@ jobs:
           path: _REPORTS
 
   tests_devmaster:
-    name: Test against ScopeSim main
+    name: Test against ScopeSim main or PR branch
     runs-on: ${{ matrix.os }}
     # Run if our target is IRDB dev_master, or when this is ran manually.
     if: github.base_ref == 'dev_master' || github.base_ref == ''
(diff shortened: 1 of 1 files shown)


teutoburg:
Only took three attempts 💪

hugobuddel (approved):
Great, thanks.

This seems much harder in github than in gitlab.

This is my 'solution' to align METIS_Pipeline and METIS_Pipeline_Test_Data https://github.com/AstarVienna/METIS_Pipeline/pull/104

teutoburg:
> This is my 'solution' to align METIS_Pipeline and METIS_Pipeline_Test_Data https://github.com/AstarVienna/METIS_Pipeline/pull/104

Yes, I saw that and took inspiration from it, but couldn't make that work with the pip install from git in the case of no matching branch. Perhaps it can be done though. For the reverse operation (IRDB inside Scopesim tests), that solution will work, because it's a straightup clone (e.g. the notebook tests), and I'm planning to use that there.

teutoburg:
Aha, gotcha: https://github.com/AstarVienna/irdb/actions/runs/13472564113/job/37647845619

When triggered by anything other than a PR (e.g. a push), the branch checking step is skipped (which was intentional, as this only has relevance for PRs). However, this means that the branch name is never set to the default (`main`), so the pip install in the next step fails. I'll fix it by moving the line that sets the default in a separate step before, that's always executed...

PR #212, opened 2025-03-04
Oc/metis wcu

oczoske:
Continuation of https://github.com/AstarVienna/irdb/pull/205 after branch has been renamed.


oczoske:
@astronomyk `TestFileStructureOfPackages::test_default_yaml_contains_packages_list` fails because the new `METIS/default.yaml` does not have `packages` and `yamls` at the same level. `packages` is defined at the top level, but the `yamls` are now specified separately for each mode. The primary reason for the test to fail is that `yaml_dict["yamls"]` does not exist. The purpose of the test seems to be to check that each of the package yamls (`{pkg_name}.yaml`) is used in `default.yaml`; the generalisation for the test would be to check that `{pkg_name}.yaml` is used _somewhere_ in `default.yaml`. But if that fails isn't all that we learn that `{pkg_name}` could be deleted?

METIS/default.yaml
```diff
@@ -16,16 +16,17 @@ changes:
...
 packages:
   - Armazones
   - ELT
   - METIS
 
-yamls:
-  - Armazones.yaml
-  - ELT.yaml                # overrides below
-  - METIS.yaml
+#yamls:
```
teutoburg, line 26:
```suggestion
yamls: []
```

oczoske:
I don't know what the status is. E-SPE-UZK-MET-1015 mentions only one grid mask and lists IMG-N among the "Beneficary subsystems". METIS-5483 requires two grid masks for 2.9 um and 7.5 um, respectively. We could make up a mask file in the irdb (same layout, larger holes?), or METIS_Simulations could use the existing one.

PR #215, opened 2025-03-13

teutoburg:
Changed:

- LFOA - tests only
- METIS (WCU et al.)
- MICADO - docs only
- WFC3 - tests only


teutoburg:
The new WCU notebook fails with:
`IsADirectoryError: [Errno 21] Is a directory: 'inst_pkgs/METIS/docs/example_notebooks/inst_pkgs'`

I'll try to fix this and then recreate the PR...

PR #216, opened 2025-03-13
Fix new WCU notebook to run in CI

teutoburg:
It seems we need to have the "../../../" path enabled by default and the `download_packages` disabled, otherwise it won't work in the CI (and probably on RTD).

### DIFF
--- a/METIS/docs/example_notebooks/METIS_WCU.ipynb
+++ b/METIS/docs/example_notebooks/METIS_WCU.ipynb
@@ -24,7 +24,7 @@
     "The most basic way to run this mode is as follows\n",
     "```python\n",
     "import scopesim as sim\n",
-    "cmd = sim.UserCommands(use_instrument=\"METIS\", set_modes=[\"wcu_lss_m\"]\n",
+    "cmd = sim.UserCommands(use_instrument=\"METIS\", set_modes=[\"wcu_lss_m\"])\n",
     "metis = sim.OpticalTrain(cmd)\n",
     "metis.observe()\n",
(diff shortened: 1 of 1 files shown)


teutoburg:
Notebook now runs in CI, see https://github.com/AstarVienna/irdb/actions/runs/13840115315/job/38725181283

hugobuddel:
The download is not necessary because the notebook is already *in* the IRDB. So the notebook should use the IRDB that it is in. It does not make sense to download an older version from the internet.

hugobuddel:
But the `sim.rc.__config__["!SIM.file.local_packages_path"] = "../../../"` should not be necessary, because there is a symlink `inst_pkgs` that points to `../../..`

teutoburg:
Agree (without trying if that actually works everywhere (GitHub Action, RTD, my local Windoof machine, ...)). This also applies to many (all?) of the other notebooks in this directory, that also manually set the "../../../" path. If we change this, we should change it everywhere.

hugobuddel:
Not all IRDB packages have this symlink though.

And there is this kludge that I wrote to make it work on Windoof machines: https://github.com/AstarVienna/ScopeSim/blob/bbaec53f3ade0adb4355ddc591d5c51daf6d4899/scopesim/commands/user_commands.py#L458

Sometimes we get errors related to that workaround, so it is not a foolproof solution, hence me not propagating it to the other packages.

Making the IRDB pip packages would probably help.

Feel free to improve the situation in any way you'd like though, I just wanted to provide some context.

PR #219, opened 2025-03-14
Override !TEL.temperature in WCU mode

### LINKED ISSUE #218: METIS WCU mode requires atmosphere and telescope
Simply running

```python
import scopesim as sim
sim.download_packages(["METIS"])
metis = sim.OpticalTrain(cmd)
metis.observe()
```

as per the example notebook, in a fresh environment, results in

```python
ValueError: !TEL.temperature was not found in rc.__currsys__
```

Attempting to solve this via `sim.download_packages(["ELT"])` still produces `ValueError: !ATMO.temperature was not found in rc.__currsys__`. Only when also including `sim.download_packages(["Armazones"])`, will it run as expected.

The culprit seems to be a surface list in the CFO, that does itself have a temperature (a number, not a reference!), but when trying to calculate the emission, it still tries to access the telescope temperature.

I don't think this is an issue with the WCU mode itself, hence no `instrument specific` label, but with ScopeSim as a whole. Perhaps I should have put the issue in ScopeSim therefore...

teutoburg:
This eliminates the need to load the ELT and Armazones packages when using METIS in one of the WCU modes. I also tested switching back and forth between e.g. "img_lm" and "wcu_img_lm" modes and !TEL.temperature does indeed switch as it should.

Fixes #218.

### DIFF
--- a/METIS/METIS_WCU.yaml
+++ b/METIS/METIS_WCU.yaml
@@ -44,3 +44,4 @@ alias: TEL
 properties:
   #area: 975.23478998
   area: 1354.593    #### TEST, needed to agree with RvB's calculation
+  temperature: "!WCU.temperature"  # Overrides !ATMO.temperature


oczoske (approved):
Perfect.

PR #220, opened 2025-03-14
Run badge report tests separately

teutoburg:
No need to run those in the whole matrix, also solves the problem of multiple artifact uploads without having to use extra flags for certain matrix combination, which was the previous stopgap solution.


hugobuddel (approved):
Good

PR #224, opened 2025-03-24
Use ExposureOutput in METIS

oczoske:
Use `ExposureOutput` in img_lm (to be expanded to other modes), see https://github.com/AstarVienna/ScopeSim/pull/603

### DIFF
--- a/METIS/METIS_DET_IMG_N_Aquarius.yaml
+++ b/METIS/METIS_DET_IMG_N_Aquarius.yaml
@@ -98,6 +98,13 @@ effects:
     kwargs:
       mode: average
 
+  - name: ad_conversion
+    description: Apply gain and convert electron count into integers
+    class: ADConversion
+    kwargs:
+      dtype: uint16
+      gain: "!DET.gain"
+
   - name: chop_nod
     descriptions: chopping and nodding
     class: ChopNodCombiner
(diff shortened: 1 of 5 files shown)


oczoske:
I have added ExposureOutput in all modes now. For `img_n` there is a question of order. I have now put averaging and quantisation (soonish to be replaced by adconversion) before chopnod, but this isn't correct either: already chopped images cannot be represented by uint16 (unless one uses BZERO).

oczoske:
Trying to sum up where we stand:
- tests are green, after fixing a stupid oversight
- The METIS example notebooks need to be checked and revised, because they do not expect AD_Conversion and with SummedExposure are not unlikely to saturate the detector output.
- The switch from SummedExposure to averaged exposure may be unexpected by some users. It shouldn't affect signal-to-noise ratio and therefore won't change "scientific" results.

METIS/METIS_DET_IMG_N_Aquarius.yaml
```diff
@@ -92,6 +92,18 @@ effects:
...
+    description: Return average or sum over NDIT subexposures
+    class: ExposureOutput
+    kwargs:
+      mode: average
+
+  - name: quantization
+    description: Turn photon count into integers
+    class: Quantization
+    kwargs:
+      dtype: uint16
```
teutoburg, line 105: Resurrection of `Quantization`?

oczoske: Oh, that. The age of Aquarius is over, so we could actually remove that file. Let's carry it around for a bit longer, without quantization.

PR #227, opened 2025-04-02
Move TEL header keywords in METIS to ELT

astronomyk:
Together with @JenniferKarr and @hugobuddel we extracted the offending references in the METIS common FITS keywords file to telescope effects. These were causing METIS simulations to fail when generating WCU mode observations, because there is (quite obviously) no telescope involved in the process. 

The Gist of this PR is that TEL keywords are now contained in the ELT IRDB package.

### DIFF
--- a/ELT/ELT.yaml
+++ b/ELT/ELT.yaml
@@ -21,6 +21,13 @@ effects :
     kwargs :
       filename : LIST_ELT_combined.tbl
 
+  - name : telescope_fits_keywords
+    description : FITS keywords specific to the telescope
+    class : ExtraFitsKeywords
+    include : True
+    kwargs :
+      filename: FITS_telescope_keywords.yaml
+
 #  - name : scope_vibration
 #    description : residual vibration of telescope
 #    class : Vibration
--- a/ELT/FITS_telescope_keywords.yaml
+++ b/ELT/FITS_telescope_keywords.yaml
@@ -1,0 +1,11 @@
+- ext_type: PrimaryHDU
+  keywords:
+    HIERARCH:
+      WISE:
+        TEL:
+          NAME: "ELT"
+          TELESCOP: "ELT"
+          TER:
+            FILENAME: "!TEL.ter_curve.filename"
+          REF:
+            FILENAME: "#telescope_reflection.filename"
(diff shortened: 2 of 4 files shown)


teutoburg:
Regarding the git issue, a quick look into the logs reveals that the branch wasn't based off of `dev_master` but `master`. I can quickly fix this, but that involves a force push, so I'm requesting @astronomyk's approval for that, as it's his branch.

teutoburg:
I was able to reproduce it myself with ScopeSim installed from `main` (v0.9.3a1) using the following minimum example:

```python
import scopesim as sim
sim.download_packages("METIS")
opt = sim.OpticalTrain(cmd)
opt.readout()
```

which throws an exception on the final line that does indeed originate in a FITS header effect.

astronomyk:
@teutoburg you have my permission to force push on my branch :+1:

Sorry for formatting - on my phone atm.

The offending line is in METIS/headers/FITS_common-keywords.yaml, way down the bottom in the WISE TEL section.  One of the entries has a #-string reference to, I think,

#telescope_reflection.filename.

Because in the METIS WCU mode we exclude the telescope package, there are no telescope effects in the OpticalTrain to field the relevant #-string call. 

Our solution was to extract the telescope items for extra FITS header keyword into the Telescope IRDB package, as that seems the logical place to keep them anyway

teutoburg:
> Because in the METIS WCU mode we exclude the telescope package, there are no telescope effects in the OpticalTrain to field the relevant #-string call.

Yeah, it also occurs generally if only the METIS package is installed (which should otherwise be sufficient for the WCU modes).

Also, @oczoske clearly stated in `METIS_WCU.ipynb`:

> To make `readout `work, we currently have to turn off the effect that adds keywords to the resulting fits file. Including it would cause failure as it expects keywords related to the telescope, which are not available in WCU mode.
> 
> ```python
> metis['common_fits_keywords'].include = False`
> ```

So if this PR fixes that, the notebook should really be updated accordingly in the same PR as well...

teutoburg:
METIS notebook test also passed: https://github.com/AstarVienna/irdb/actions/runs/14229069153/attempts/1#summary-39875480888

teutoburg (approved):

PR #228, opened 2025-04-15
WCU FITS header keywords

oczoske:
The METIS WCU mode gets an instance of `ExtraFitsKeywords` which writes the main WCU parameters to the FITS header. The sequence
```
metis = sim.OpticalTrain(cmd)
metis['wcu_source'].set_lamp("laser")
metis.observe()
hdul = metis.readout()[0]
```
produces
```
HIERARCH ESO INS OPTI17 ID = 'WCU periscope'                                    
HIERARCH ESO INS OPTI17 NAME = 'IN      '                                       
HIERARCH ESO INS OPTI18 ID = 'WCU PP1 masks'                                    
HIERARCH ESO INS OPTI18 NAME = 'open    '                                       
HIERARCH ESO INS OPTI19 ID = 'WCU BB exit aperture'                             
HIERARCH ESO INS OPTI19 NAME = 1.0                                              
HIERARCH ESO INS OPTI20 ID = 'WCU FP2.1 mask'                                   
HIERARCH ESO INS OPTI20 NAME = 'open    '                                       
HIERARCH ESO INS WCU_LAMP ID = 'WCU lamp'                                       
HIERARCH ESO INS WCU_LAMP NAME = 'laser   '                                     
HIERARCH ESO SEQ WCU_BB_TEMP = 1000                   
```
The `OPTI` and `SEQ WCU_BB_TEMP` keywords were taken from the template manual (`ID`s made up, `NAME` does not seem quite right for the values); `INS WCU_LAMP` is missing from the templates, so I made those up.


oczoske:
`FITS_wcu_keywords.yaml` caused `test_yaml_entries_are_dicts` to complain as it contains a list at top level. The same is true for `FITS_common_keywords.yaml` but since that is in a subdirectory it is not tested. Rather than question the validity of the test (which is geared towards the instrument definition yamls), all FITS definition yamls should from now on be collected in `headers/`.

METIS/headers/FITS_wcu_keywords.yaml
```diff
@@ -0,0 +1,23 @@
...
+            ID: "WCU BB exit aperture"
+            NAME: "#wcu_source.bb_aperture"
+          OPTI20:               # WCU FP2.1 mask wheel
+            ID: "WCU FP2.1 mask"
+            NAME: "#wcu_source.current_fpmask"
+          WCU_LAMP:             # WCU lamp TODO: made up keyword
+            ID: "WCU lamp"
+            NAME: "#wcu_source.current_lamp"
+        SEQ:
+          WCU_BB_TEMP: "#wcu_source.bb_temp"
```
hugobuddel, line 23: It is recommended to use 8 characters for the parameter names in ESO-044156, the Data Interface Control Document. From v7 section 4.4.1.3 "Parameters":

>  In order to keep the size required by the complete hierarchy to a minimum, it is recommended to use names not exceeding 8 characters. 

I'm fine with longer names, but good to do that consciously.

METIS/headers/FITS_wcu_keywords.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+    HIERARCH:
+      ESO:
+        INS:
+          OPTI17:               # WCU periscope arm
+            ID: "WCU periscope"
+            NAME: "IN"
+          OPTI18:               # WCU PP1 masks
+            ID: "WCU PP1 masks"
+            NAME: "#pupil_masks.current_mask"
```
hugobuddel, line 12: Are you sure this is the proper use of `ID` and `NAME`?

From the filters I learned that `NAME` is a human-readable name for the filter, e.g. "SLOAN_r", which is similar to what you have, but that `ID` is a reference to a specific physical piece of hardware. E.g. a numerical value.

That is, say we have the same filter fabricated twice, and put one in the instrument and have the other as spare. If the one in the instrument will at some point be replaced by the spare, then the `NAME` stays the same, but the `ID` will change.

If that also applies to the other OPTI instruments, then I would expect something different in the `ID` column.

At least, that is my understanding, which could be wrong.

hugobuddel (approved):
Approved! Thanks @oczoske .

See above (or below?) for some questions.

Perhaps we should include some tests for the header keywords? In my experience this is something that easily breaks... But I'll leave that decision to you.

METIS/headers/FITS_wcu_keywords.yaml
oczoske: Oh, I though `ID` was the identifier for the optical element (i.e. `OPTI18`), not for the mask itself. But what you say makes sense. Let me check some other instruments.

METIS/headers/FITS_wcu_keywords.yaml
hugobuddel: Thanks for the link. Yes, it's better to stay consistent with the template manual.

METIS/headers/FITS_wcu_keywords.yaml
hugobuddel: `DESC` sounds good.

PR #234, opened 2025-05-13

teutoburg:
### ELT

- 🗂️ header keywords

### METIS

- 🗂️ header keywords
- 🚧 `Quantization` ➡️  `ADConversion`
  ⚠️ requires ScopeSim v0.10, not yet released!
- 📓 docs
- ✅ tests


teutoburg:
Tests against ScopeSim release fail because `ADConversion` is not yet released. I'm fine with ignoring that for now, although as mentioned I think it would be useful to add a `needs_scopesim: 0.10.0` key in the package metadata. But that's for later...

oczoske:
How does this avoid breaking people's scopesim runs if the instrument package asks for a `ADConversion` effect that does not exist in the current (pip) scopesim version?

teutoburg:
> How does this avoid breaking people's scopesim runs if the instrument package asks for a `ADConversion` effect that does not exist in the current (pip) scopesim version?

It doesn't.

After a brief offline discussion, we decided that this isn't ready yet, and to rather create dev version tags and package those. This release should be made only _after_ ScopeSim v0.10.0 is released (probably next week).

PR #236, opened 2025-05-25
LSF effect description for mode `lms_cube`

oczoske:
Implementing LSF effect (https://github.com/AstarVienna/ScopeSim/pull/670) for METIS `lms_cube`

### DIFF
--- a/METIS/default.yaml
+++ b/METIS/default.yaml
@@ -6,7 +6,7 @@ alias: OBS
 name: METIS_default_configuration
 description: default parameters needed for a METIS simulation
 status: development
-needs_scopesim: "0.9"
+needs_scopesim: "v0.9"
 date_modified: 2024-09-04
 changes:
   - 2021-12-16 (OC) chopnod defaults to perpendicular
@@ -366,7 +366,7 @@ mode_yamls:
     name: lms_cube
     description: "METIS LM-band integral-field spectroscopy, nominal mode, cube output"
     status: experimental
-    needs_scopesim: "0.10.0a1"
+    needs_scopesim: "v0.10.0a1"
     yamls:
       - Armazones.yaml
       - ELT.yaml


teutoburg:
This is now a feature branch based on another feature branch. Which is perfectly fine, but has to be treated with some care. I created some diagrams below to illustrate the situation (commit IDs and number of commits is simplified for the example).

## Current situation
```mermaid
%%{init: {'gitGraph': {'mainBranchName': 'dev_master'}} }%%
gitGraph
   commit
   branch fh/simple-ifu
   commit
   commit
   commit
   branch oc/lsf_effect
   commit
   commit
   checkout dev_master
   merge fh/simple-ifu id: "merge 1"
   merge oc/lsf_effect id: "merge 2"
```
This is how it would look like if we first merged #233 and then this (the other way around wouldn't make sense). The issue here is that this PR here will still show the commits from `fh/simple-ifu` _even after that branch is merged into_ `dev_master`. That seems a bit "unclean" to me.

## Alternative
```mermaid
%%{init: {'gitGraph': {'mainBranchName': 'dev_master'}} }%%
gitGraph
   commit
   branch fh/simple-ifu
   commit
   commit
   commit
   branch oc/lsf_effect
   commit
   commit
   checkout fh/simple-ifu
   merge oc/lsf_effect id: "merge 1"
   checkout dev_master
   merge fh/simple-ifu id: "merge 2"
```
The other option, which I'd much prefer, is to first merge `oc/lsf_effect` into its parent `fh/simple-ifu`. Then that branch would receive the commits from here, and could be merged into `dev_master` at once. The PR here should then only show the commits done _after_ branching off of `fh/simple-ifu` in the diff, which seems much more representative of what actually happened.

### How to?
_Should_ be as simple as **changing the base for this PR** to `fh/simple-ifu` instead of the default `dev_master`. I haven't done that many times, but I'm fairly sure it'll work. If not, we can always change it back, that doesn't destroy anything.

## Alternative alternatives
### Merging `dev_master` back into `oc/lsf_effect` after merging `fh/simple-ifu`
While that should also avoid the duplication of commtis, it seems like some unnecessary criss-cross merging and also I'm not entirely sure how the commits would be resolved. Better avoid.
### Rebasing
Would certainly work, but involves force pushing, which is not so great if the PRs already exist.
### Just merging this PR
Would also work, i.e. closing #233 unmerged and then only merging this PR, meaning we pretend `fh/simple-ifu` never existed and all the commits were just done on `oc/lsf_effect`. But that seems like "rewriting history" and having the closed PR in there might be confusing in the future. Better also avoid.

teutoburg (approved):
Changed the base as described in previous comment. To my relief, it now actually looks as I expected. Go ahead and merge 👍

PR #241, opened 2025-06-15
Restructure CI workflows

teutoburg:
The current setup is very clumsy and hacky, and in several details doesn't do what we need, e.g. running notebook tests from "sister branches". I also ran into the situation many times that I wanted to run them also for a PR or just for a change in ScopeSim (without a corresponding PR in the IRDB), to see if I'm about to break anything.

For all those reasons, I "pulled apart" that one large workflow file into multiple separate, clean and simple workflows that rely on each other. I also added an attempt of an explanation in the form of `CI_strategy.md`, which should hopefully (togehter with comments in the workflow files) serve as documentation for all of this, as well as a place to share ideas for further improvement, as some aspects of all of this are not yet final. Especially those badge reports are still not really used, just created. I do have a few things in mind how to do that, but that'll be for later.

### DIFF
--- a/CI_strategy.md
+++ b/CI_strategy.md
@@ -9,15 +9,16 @@ A brief guide to how the CI works here, because it's getting rather complex alre
 * Run normal tests
 * Run notebook tests when ready for review
 
-Both using corresponding branches in ScopeSim and ScopeSim_Templates if available.
+Both using corresponding branches (with the same name) in ScopeSim and ScopeSim_Templates if available.
 
 ### Into master
 
 * Run normal tests
 * Run notebook tests
 
-Both using released version of ScopeSim and ScopeSim_Templates, because once it's in master, it can and should become a new package relase and those need to work with the released versions.
-For anything that requires changes in more than one place, release the IRDB part last.
+Both using released version of ScopeSim and ScopeSim_Templates, because once it's in master, it can and should become a new IRDB package release and those need to work with the released versions of the software.
+For anything that requires changes in more than one place, please release the IRDB part last.
+This means that e.g. the new version of ScopeSim should be compatible with both the old and the new version of the IRDB.
 
 ## Pushes
 


teutoburg:
Now of course, the now-new workflows don't trigger until they've been merged in, so a bit tricky to test this...

CI_strategy.md
```diff
@@ -0,0 +1,72 @@
...
+
+Both using corresponding branches in ScopeSim and ScopeSim_Templates if available.
+
+### Into master
+
+* Run normal tests
+* Run notebook tests
+
+Both using released version of ScopeSim and ScopeSim_Templates, because once it's in master, it can and should become a new package relase and those need to work with the released versions.
+For anything that requires changes in more than one place, release the IRDB part last.
```
hugobuddel, line 20: Perhaps add something like:
```suggestion
For anything that requires changes in more than one place, please release the IRDB part last.
This means that e.g. the new version of ScopeSim should be compatible with both the old and the new version of the IRDB.
```

CI_strategy.md
```diff
@@ -0,0 +1,72 @@
...
+* Run normal tests
+* Run notebook tests
+
+Both using main branches in ScopeSim and ScopeSim_Templates.
+Useful to catch cases where PR base was out-of-date and PR CI passed but now might fail after merge.
+Also useful to have up-to-date badge, see also schedule.
+
+### To master
+
+Tests?? Why actually?
```
hugobuddel, line 35: Maybe not, but doesn't really matter much. I would suggest to tests master and dev_master on schedule. Then it should not be necessary to test master on push. But then again, it shouldn't cause problems either.

CI_strategy.md
```diff
@@ -0,0 +1,72 @@
...
+
+Maybe at this point also re-run notebooks with output (the ones too long for RTD) and commit the changes, so that RTD shows the current state?
```
hugobuddel, line 37: Yeah running the notebooks with output and committing those sounds good.

CI_strategy.md
teutoburg: Or rather an exclamation mark at the end 😛

CI_strategy.md
teutoburg: In the end I thought that the case of "another PR changed the state in between so the status of the PR checks are not representative of what we'll end up with after the merge" should really not occur for PRs onto master. Sidenote: I also learned in the meantime that there are "merge queues" that might solve issues like that, but I don't fully understand them yet. But idk, I have to keep thinking about this for a little longer to be definitive.

carmeloarci:
I'm using the released version of ScopeSim,
focusing on MICADO simulation

!pip install scopesim scopesim_templates

import scopesim 
import scopesim_templates 
scopesim.download_packages(["Armazones", "ELT", "MICADO", "MORFEO"], release="github@master")

In this way, it happens that I'm downloading the 2023 release, similarly to
scopesim.download_packages(["Armazones", "ELT", "MICADO", "MORFEO"])

Otherwise, calls like the following point to 2017 data:
filter_name="elt/micado/I-long"
stars_i = scopesim_templates .stars(filter_name,mags,spec_arr,x,y)
# J
filter_name="elt/micado/J"
stars_j = scopesim_templates .stars(filter_name,mags,spec_arr,x,y)
filter_name="elt/micado/H"
stars_h = scopesim_templates .stars(filter_name,mags,spec_arr,x,y)
filter_name="elt/micado/Ks"
stars_ks = scopesim_templates .stars(filter_name,mags,spec_arr,x,y)

Any suggestion to make the proper call to the MICADO 2025 filter curves?
Regards

hugobuddel:
I believe the new filter curves have not yet been merged into the IRDB.

You can follow along with the development in https://github.com/AstarVienna/irdb/pull/242

teutoburg:
Yes, that and also there's a bit of a bug with version number when installing from source (e.g. via `release="github@master"`). That's no to say the wrong version gets installed, but the version _displayed_ can be wrong.

PR #247, opened 2025-06-21
METIS filters cut to reasonable wavelength range

oczoske:
The new METIS filter transmission curves cover broad wavelength ranges from 2 to 5.87 um (LM) and 2 to 15.8 um (N). This caused Scopesim to try to create FOVs covering this (or rather the full instrumental wavelength range). The filter curves are now cut to the wavelength range where the transmission is significantly different from 0 (typically >  0.001).

This should be sufficient to fix https://github.com/AstarVienna/ScopeSim/issues/705.

METIS_filters_cut


hugobuddel:
Perhaps it would be good to add a leading and trailing 0-transmission line. I forgot why this was useful though.

oczoske:
> Perhaps it would be good to add a leading and trailing 0-transmission line. I forgot why this was useful though.

Many moons ago spectra were extrapolated using the available edge values, which caused huge spurious throughput. Oh bugger, that is still the case (damn synphot). I don't think that is relevant, though, as the available wavelength range of the filter table should set the FOV's wavelength range.  To be checked.

teutoburg (approved):
This also reduces the overall size of the METIS package (still larger than in previous release for obvious reasons, but not as dramatic), without loosing any actual information really, just a bit of noise.

PR #248, opened 2025-06-26
Add `ExposureOutput` and `ADConversion` to MICADO

### LINKED ISSUE #238: Add `ExposureOutput` to MICADO

teutoburg:
Keeping the mode for `ExposureOutput` as "sum" for now to hopefully not break the notebooks. This can be changed later.

How could the `ADConversion` get the gain value from `FPA_metis_img_lm_layout.dat`?

Also, some indentation fixing and spelling (When was the last time someone looked at this file? What's with all the commented-out stuff?).

Closes #238.

### DIFF
--- a/MICADO/MICADO_H4RG.yaml
+++ b/MICADO/MICADO_H4RG.yaml
@@ -1,29 +1,30 @@
 ### H4RG DETECTOR
-object : detector
-alias : DET
-name : MICADO_DET
-description : A set of 9 H4RG detectors
+object: detector
+alias: DET
+name: MICADO_DET
+description: A set of 9 H4RG detectors
 
-properties :
-    image_plane_id : 0
-    temperature : -230
-    dit : "!OBS.dit"
-    ndit : "!OBS.ndit"
+properties:
+    image_plane_id: 0
+    temperature: -230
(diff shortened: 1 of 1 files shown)


oczoske:
> How could the `ADConversion` get the gain value from `FPA_metis_img_lm_layout.dat`?

 metis or micado? For METIS, all gain values are defined in detector yamls. For MICADO, the solution could be like the one for METIS_IFU, although we will have to be extra careful with the detector ids once we get individual gain values for the detectors. Getting them from the FPA layout would be safer in that respect, but that doesn't work (out of the box) for detectors with several modes.

teutoburg:
Of course I mean `FPA_array_layout.dat` for MCIADO. How on Earth did I manage that copy-paste mixup and not see it, despite reading the whole thing at least thrice? And now this is even in the commit message, oh dear...

oczoske (approved):
Fine for this week's release, but to be completed for the next one. 
- we should work towards having individual characteristics for the detectors, not just gain but also dark current, read noise,... DarkCurrent and ADConversion can read dictionaries in the yamls, not sure about ReadoutNoise.
- extension headers should be filled with these parameters

PR #251, opened 2025-06-27
Fix METIS notebooks

teutoburg:
In preparation of release.

### DIFF
--- a/METIS/headers/FITS_det_ifu_keywords.yaml
+++ b/METIS/headers/FITS_det_ifu_keywords.yaml
@@ -10,7 +10,7 @@
 
 - ext_number: [1]
   keywords:
-
+    BUNIT:            "ADU"
     HIERARCH:
       ESO:
         DET3:
@@ -24,7 +24,7 @@
 
 - ext_number: [2]
   keywords:
-
+    BUNIT:            "ADU"
     HIERARCH:
       ESO:
         DET3:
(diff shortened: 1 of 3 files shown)


METIS/headers/FITS_det_ifu_keywords.yaml
teutoburg, line 1: Ultimately I think `BUNIT` should be changed by each effect listed in [the new flowchart](https://scopesim.readthedocs.io/en/stable/faqs/scopesim_units_flow.html), and then would just stay in the header with the output. That way the `BUNIT` in the output would be the correct one, even if some effect like `ADConversion` is switched off. "Forcing" it to ADU through these somewhat hidden FITS keyword effects might lead to some confusion (units in ScopeSim are already confusing enough for new users). But we can change that later in the course of AstarVienna/ScopeSim#662, this now is alright as a quick fix...

teutoburg: I don't think _every_ FITS keyword in the output needs to be set via one of those FITS header effects. While I agree that we should get rid of `OpticalTrain.write_header()` (or at the very least severely limit what it does), I think some of the more "basic" keywords can still be set through other means. Things like the WCS keywords for example. Anything that uses a complex (i.e. HIERACH) keyword should go through one of the FITS header effects though.

hugobuddel (approved):
I didn't yet run all the notebooks, but this looks fine. Thanks for taking care of them @oczoske

PR #252, opened 2025-06-30
Remove redundant !!float in yamls, fix others

teutoburg:
Since forever, anything that has a decimal point and an explicit plus or minus in the exponent is converted to a float anyway, no need for this ugly tag. Also, "0." is converted to `0.0` anyway.

While in the process, I standardized on a lowercase "e". Both works fine and we had both before, but why not.

### DIFF
--- a/HAWKI/HAWKI.yaml
+++ b/HAWKI/HAWKI.yaml
@@ -51,8 +51,7 @@ effects:
               - NB2090
           filename_format: "!INST.filter_file_format"
           current_filter: "!OBS.filter_name"
-          minimum_throughput: !!float 0.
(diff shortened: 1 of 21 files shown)


oczoske (approved):
I cannot test it myself right now, but I trust it doesn't break anything.

hugobuddel:
Well, not since 'forever', the Python yaml parser had problems with parsing floats. (E.g. when the dot was or was not missing and/or whether there was an e or an E, I forgot the details.)

It probably is fine now, since https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell does not mention accidentally parsing floats as strings (only the other way around).

PR #253, opened 2025-06-30

teutoburg:


teutoburg:
I also now added the badge report files of the most recent run here (which triggers another run of those tests, that still needs tweaking...), as those haven't been updated in forever, and this merge into `master` seems like the right time to commit those.

hugobuddel (approved):
Let's go!

PR #265, opened 2025-09-17
MICADO LSS updates to example notebook and psf generation script

janusbrink:
Updated the MICADO SPEC example notebook to work better with the latest ScopeSim behaviour (e.g. PSF handling). Also changed the example a bit to start with the short 3" slit that can be simulated quicker. PSF interpolation is set to off by default to speed things up as well.

The RO yaml file was updated to allow easier setting of custom PSFs from within a notebook via the user commands.

The PSF generation script was updated as well to fix the output FITS WCS headers.


oczoske (approved):
I would suggest renaming `utils.py` to a name that indicates what it does. A bit more effort would be to make `N` and `offset` script parameters. Maybe move into directory `code` (if meant to be archived) or into new directory `utils` or `util_scripts` or something like that (if meant to be a tool for users).

PR #267, opened 2025-09-19
Added MICADO rectify traces example notebook

janusbrink:


oczoske:
I just tried it with a rectified FITS files - the wavelength axis looks a bit silly because astropy converts spectral WCS always to meters...

PR #270, opened 2025-09-25
FITS keyword values Boolean rather than string

### LINKED ISSUE #269: Boolean FITS header values are coded as strings rather than true Booleans
Comment from Lars Lundin. "The boolean header keywords in our simulated data are strings "T" and "F" instead of boolean T, F."

An example would be `HIERARCH ESO DET3 CUBE MODE = 'F       '`.  A counterexample is
```
HIERARCH ESO INS OPTI3 DESC = 'Slit wheel'
HIERARCH ESO INS OPTI3 NAME = F
```

The problem arises, when the yamls set the value explicitely as `"F"`/`"T"`; it works fine when Boolean values `false`/`true` (and presumably `yes`/`no`) are used.

oczoske:
Fixes #269 . All obvious occurrences of `"T"`/`"F"` in METIS and MICADO yamls have been replaced by `true`/`false`. This may have missed some indirect/imported occurrences - we need to keep an eye on this.

### DIFF
--- a/METIS/headers/FITS_common_keywords.yaml
+++ b/METIS/headers/FITS_common_keywords.yaml
@@ -81,7 +81,7 @@
             NAME: "TODO"
           CUBE:
             # TODO: Allow cube mode in ScopeSim.
-            MODE: "F"
+            MODE: false
         INS:
           # Keywords from the instrument subsystem.
           #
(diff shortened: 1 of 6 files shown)


gotten:
I inspected a header from a Scopesim output HDU with this IRDB branch and can see the INHERIT keyword is now the boolean T in the header and is evaluated as True when printed in the command line. So I am OK with merging this.

JenniferKarr:
Looks good to me

PR #272, opened 2025-09-28
IPC effect for METIS H2RG detectors, demo notebook

oczoske:
IPC effect has been added for the H2RG detectors in METIS (IMG_LM, IFU). Could also be done for MICADO (of interest to the astrometry group?).

Requires https://github.com/AstarVienna/ScopeSim/pull/806

### DIFF
--- a/METIS/default.yaml
+++ b/METIS/default.yaml
@@ -6,8 +6,8 @@ alias: OBS
 name: METIS_default_configuration
 description: default parameters needed for a METIS simulation
 status: development
-needs_scopesim: "v0.10.0b4"
-date_modified: 2025-06-27
+needs_scopesim: "v0.11.1a1"
+date_modified: 2025-09-29
 changes:
   - 2021-12-16 (OC) chopnod defaults to perpendicular
   - 2021-12-16 (OC) default slits renamed
@@ -19,6 +19,7 @@ changes:
   - 2022-03-14 (OC) use single PSF file
   - 2024-09-04 (OC) rearrange to include WCU modes
   - 2025-06-27 (FH) version bumps and status updates
+  - 2025-09-29 (OC) version bump
 
 packages:
   - Armazones


METIS/docs/example_notebooks/demos/demo_inter_pixel_capacitance.ipynb
teutoburg, line 1: Didn't we agree on saving notebooks without output to produce shorter diffs, except for the ones that are very resource-heavy in execution (which I assume / hope this one isn't)? In any case, I need to finalize my RTD refactor here, so that all (small) notebooks are run during the docs build...

teutoburg (changes requested):
To ensure this is run with a version of ScopeSim that actually has the new effect (once we're no longer in matching branches), I suggest updating the `needs_scopesim` key to `0.11.1a1` in `default.yaml` (can't add it as a suggestion because that file wasn't touched here), and _before that_, bump accordingly with `poetry version prerelease` in the ScopeSim branch to match. This is a good case to actually try these things 🙂

METIS/docs/example_notebooks/demos/demo_inter_pixel_capacitance.ipynb
oczoske: I thought I'd cleared the output? Obviously I thought wrong.

teutoburg (approved):

METIS/docs/example_notebooks/demos/demo_inter_pixel_capacitance.ipynb
hugobuddel: We would need to squash the branches for the removal of the output in 5d73929c0ea09b7ec2d6f450954c3c176689fd93 to be useful.

If we don't squash, then the output will still be in the git history taking up space. And in that case, we might as well have kept the output, as the space is lost anyway. Squashing will ensure the figures stay in this branch and will never exist in the history of the main branch. (And this branch can then be deleted.)

But I personally think that these figures are probably small enough that it would be fine to just include them.

PR #279, opened 2025-10-22
include fine grid in grid_lm mask; delete grid_lms mask

oczoske:
This PR merges the LMS grid (5x5 pattern at the centre of the mask) into the LM grid. There is therefore a single mask `fp_mask_grid_lm.dat` that is to be used for simulations of both the LM imager/spectrograph and the LMS IFU. The mask `fp_mask_grid_lms.dat` has been removed.

The first image plots the full mask (arcsec, hole diameters not to scale); the second is an IMG_LM detector image showing the central part of the mask (pixels).
lm_grid_mask_full
wcu_grid_lm


oczoske:
As far as I can see `METIS_Simulations` only uses `grid_lm`, even for `IFU_RSRF_RAW1`. In fact the lms mask was a fairly recent addition, which doesn't seem to have been picked up yet.

PR #291, opened 2025-11-26
Default pupil masks and psfs for wcu modes

oczoske:
The WCU modes use "perfect" PSFs computed by Roy van Boekel for the various pupil masks. The PR places all pupil masks in the pupil wheel in the WCU (`METIS_WCU.yaml`). This differs from the instrument configuration where the masks are in several wheel in the CFO, the imager and the LMS. However, for the time being these PSFs are only used for AIT with the WCU. In a future version we may think about using them for on-sky observations as well.
The default configuration uses the following pupil masks and their corresponding psfs:
| mode | mask |
| --- | --- |
| wcu_img_lm | PPS-LM |
|wcu_img_n     | PPS-N |
|wcu_lss_l, _m, _n | PPS-CFO2 |
|wcu_lms | PPS-LMS |

Relates to https://github.com/AstarVienna/ScopeSim/issues/832 and https://github.com/AstarVienna/ScopeSim/pull/838


METIS/default.yaml
```diff
@@ -63,6 +63,7 @@ properties:
   ra: 0.0
   dec: 0.0
   # These defaults are overruled by the individual modes:
+  psf_name: none
```
teutoburg, line 66: What will be used as the default PSF now? Previously, it was `PSF_SCAO_9mag_06seeing.fits`, but now I don't see any actual filename as a default. Just to make sure a default METIS simulation now still uses _a_ PSF at all.

teutoburg (approved):
See comment. Otherwise go 👍 (or perhaps wait for the ScopeSim counterpart?)

METIS/default.yaml
oczoske: The sky modes all have a `psf_file` (in `default.yaml`). I'll check whether that still works.

Issue #292, opened 2025-11-27
Implement masked regions for METIS detectors


JenniferKarr:
The METIS image data needs masked regions around the edge of the detector for crosstalk correction (see section 10.2 of the DRLD)

oczoske:
Unfortunately there is no Scopesim effect yet, cf. https://github.com/AstarVienna/ScopeSim/issues/839.

The dimensions of the METIS masks are as follows:
|detector | left | right | top | bottom | note |
|--- |--- |--- |--- |--- |--- |
|IMG-LM | 64 |64 |64 |64 |[METIS-4909](https://polarion.astron.nl/polarion/#/project/METIS/workitem?id=METIS-4909) |
|IMG-N |28 |28 |28 |28 | [METIS-4910](https://polarion.astron.nl/polarion/#/project/METIS/workitem?id=METIS-4910),  E-ANR-ESO-MET-1004_1-0 Sect.4.3.4|
| LMS | 64/0 | 0/64 | 32 | 32 | [METIS-9141](https://polarion.astron.nl/polarion/#/project/METIS/workitem?id=METIS-9141) |

PR #293, opened 2025-11-27
ReferencePixelBorder for METIS IMG-LM, IMG_N and LMS

oczoske:
Implements the reference pixels as in #292 

The dimensions of the mask are recorded in the output fits headers of the image HDUs as `ESO.DETi.CHIP.REFPIX`.

### DIFF
--- a/METIS/headers/FITS_det_ifu_keywords.yaml
+++ b/METIS/headers/FITS_det_ifu_keywords.yaml
@@ -3,7 +3,7 @@
     HIERARCH:
       ESO:
         DET3:
-          DIT:        "!OBS.dit"
+          DIT:        ["!OBS.dit", "[s]"]
           NDIT:       "!OBS.ndit"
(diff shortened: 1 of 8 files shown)


oczoske:
Some of the tests are currently broken, need to check.
--- later: the reference pixel mask messed up the means and sums (over the entire detector) that were used in the test. The test now runs without the mask, as before.

METIS/headers/FITS_det_ifu_keywords.yaml
```diff
@@ -3,7 +3,7 @@
     HIERARCH:
       ESO:
         DET3:
-          DIT:        "!OBS.dit"
+          DIT:        ["!OBS.dit", "[s]"]
```
teutoburg, line 6: That just ... works? Nice!

teutoburg (approved):
👍

PR #294, opened 2025-11-28
Fix and improve package publishing infrastructure

### LINKED ISSUE #258: Update publish script cryptography
The recent paramiko update (4.0.0) broke our CI and ScopeSim_Data. This was temporarily patched in #257.

Quoting from paramiko's release notes:

> Removed support for the DSA (aka DSS) key algorithm, as it has been badly outdated and insecure for a decade or more at this point, and was recently completely removed from OpenSSH as well.
> 
> If you were still using DSA out of sheer inertia: we strongly recommend upgrading to Ed25519 (or maybe ECDSA).

I guess we should figure out where we're actually using this and then switch if we can. Might be inside `pysftp`, in which case this could be a bit more work, as their last release was in 2016...

teutoburg:
Most importantly, this fixes #258 and gets rid of the abandoned `pysftp` as a dependency.

While going through that part of the code, I also included a ton of cleanup and removed several long-deprecated and unused functionality, as well as some more mundane things like everyone's favorite, docstrings & formatting.


oczoske (approved):
Not an in-depth review (basically because I'm out of my depth here), but I didn't see anything odd or suspicious glancing through the changes.

hugobuddel (approved):
Certainly will make things better

PR #296, opened 2025-12-04
Add JWST and NIRCam packages

teutoburg:
Very basic versions of both for now, mostly because @astronomyk would like to recreate a comparison image between HST, JWST and MICADO. Some aspects like filter curves are more or less final, other like PSFs are a good initial approximation. The long wavelength imaging mode (the main one used here) currently uses a "fake" detector layout for simplification purposes, but the more accurate layout is already included.


oczoske (approved):
Looks alright. Obviously I haven't checked the numbers. Example notebooks would be nice.

PR #298, opened 2025-12-10
Add fake !ATMO.location to HST and JWST packages

teutoburg:
This avoids a non-critical error seen in `OpticalTrain.readout()`.

### DIFF
--- a/HST/HST.yaml
+++ b/HST/HST.yaml
@@ -13,3 +13,9 @@ effects :
     class : SurfaceList
     kwargs :
         filename : LIST_mirrors_HST.tbl
+
+---
+alias: ATMO
+description: "Fake !ATMO.location to avoid confusing ScopeSim"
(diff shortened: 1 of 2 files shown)


oczoske (approved):
Space is the place. Maybe next time give a little detail on where the error occurred.

teutoburg:
The error occurs at [this line](https://github.com/AstarVienna/ScopeSim/blob/158b4d26d61f757fb1c93ab7f887358a0ff405f8/scopesim/optics/optical_train.py#L506), where `!ATMO.location` is queried. But the whole function call is [wrapped in a try-except](https://github.com/AstarVienna/ScopeSim/blob/158b4d26d61f757fb1c93ab7f887358a0ff405f8/scopesim/optics/optical_train.py#L473-L478), so the error is caught. Still ugly in the logs.

PR #299, opened 2025-12-11
Fix and cleanup WFC3 package 2/2

teutoburg:
Continuation of #295.

- Moved TER (aka filter curve) files (.dat) to subfolder for less crowded structure. @oczoske will be happy to see that I used `git mv` this time 🙃
- Remove redundant FITS files that seem to have been used to create said TER files at some point. No real point in keeping those around for this package...
- There was no PSF effect of any kind present in the package. A FITS file containing field varying PSF data was available, but to use it required AstarVienna/ScopeSim#851.
- Various minor cleanup and formatting.


WFC3/default.yaml
```diff
@@ -3,54 +3,64 @@
...
 
-# default simulation parameters
-object : simulation
-alias : SIM
-name : WFC3_simulation_paramters
-description : RC simulation paramters which need to change for a WFC3
+---
+object: simulation
+alias: SIM
+name: WFC3_simulation_paramters
```
oczoske, line 56: paramters -> parameters (fixed in #302 , which will probably cause a minor conflict)

WFC3/default.yaml
```diff
@@ -3,54 +3,64 @@
...
+description: RC simulation paramters which need to change for a WFC3
```
oczoske, line 57: same

oczoske (approved):
Looks fine.

PR #301, opened 2025-12-16
Measured MICADO filter curves

oczoske:
This PR replaces a number of MICADO filter transmission curves with newly received measured data. The figure shows the new data in blue, compared to the old values in orange. The wavelength range could be restricted to the cut-off of the detectors. We're not doing that yet as we do not have measured quantum efficiency data on the actual detectors. 

MICADO_new_filters


janusbrink (approved):
Looks good - thanks. So this update only includes the as-measured updates, which is fine since the _as-designed_ values we have from Ric seems more accurate than the ones provided with the update.

janusbrink:
One slightly odd thing is that some measured values are > 100.00% (e.g. J-80K). Hope ScopeSim handles that OK...

oczoske:
Oh dear, what are these people measuring???
The affected files are
- `TC_filter_J-long.dat` 1.01434
- `TC_filter_xJ2.dat` 1.00055
- `TC_filter_J-short.dat` 1.00543
- `TC_filter_H-long.dat` 1.01938
- `TC_filter_J.dat` 1.01626
It's not dramatic.

janusbrink:
It should be fine, yes - it's just due to the measurement error level.

PR #302, opened 2026-01-02
Set all random seeds to None by default

### LINKED ISSUE #6: Set random seed to `None` by default
All default.yaml files currently set a random seed (9001). This results in identical shot noise when series of observations/readouts of the same Source are made (I noticed this in chopped images with METIS IMG_N).  I think it would be better to have random_seed=None by default and instead explain in the documentation how to set a seed should the need arise.

oczoske:
Setting a random seed leads to identical simulation results. This is sometimes useful but may lead to unwanted results, e.g. identical noise patterns when performing multiple observations of the same source. This PR changes `!SIM.random.seed` to None for those instruments that haven't had it already. Users still have the possibility to set a seed for their simulation, although documentation may not be adequate (currently restricted to  https://scopesim.readthedocs.io/en/latest/5_liners/simulation_parameters.html, I think).

Closes #6


teutoburg (approved):
Thanks for finally closing a 5 yr old issue 😃 

I took the liberty to solve the merge conflict caused by (my) #299.

PR #308, opened 2026-02-03
Move slit to MICADO.yaml; general spec mode

### LINKED ISSUE #88: [MICADO] Slit-through images
We need the possibility to simulate slit-through images with MICADO. This works if the effect `spectroscopic_slit_aperture` is moved from `MICADO_spec.yaml` to `MICADO.yaml` and `!OBS.slit_file` is set to an existing mask file. However, we have not found a way to disable the slit by default, e.g. by setting `!OBS.slit_file` to `false` or `none` or something like that. METIS uses a slit wheel effect, which is enabled in imaging but with `slit = false` as default.

oczoske:
The MICADO slit selection is moved to `MICADO.yaml` in the form of a `SlitWheel`:
- Makes it easier to introduce new slits (shifted slits, pinhole slits)
- Enables through-slit imaging.

A new mode `SPEC` is introduced, which replaces all current spectroscopic modes. The old modes `SPEC_15000x20` etc. still exist and should work as before with default settings as indicated by their name.

FITS keyword `INS.SLIT.NAME` is updated when the slit is changed. As the pipeline relies on the names being "Short", "Wide", "Long", the slit definition files are renamed accordingly:
- 15000x20 -> Long
- 15000x20_offset -> Long_offset
- 3000x16 -> Short
- 3000x16_offset -> Short_offset
- 3000x48 -> Wide
- 3000x48_offset -> Wide_offset
Again, the old modes keep their name.

A new notebook demonstrating the new slit mode has been added to `MICADO/docs/extra_notebooks`. Maybe this could be integrated with `6_basic_spectroscopy.ipynb`, but I'll leave it stand-alone for now. Notebooks in the new directory `extra_notebooks` are not automatically run by github actions.

Needs https://github.com/AstarVienna/ScopeSim/pull/875
Closes #88


teutoburg:
I suggest adding
```yaml
  sub_pixel:
    flag: True
```
to the `default.yaml` at the bottom.

While not strictly related to the rest of this PR, it seems like a useful default and doesn't appear to impact performance too badly, at least on my machine...

oczoske:
Sorry for all those commits (from rebasing dev_master). The relevant commits for this PR are the first ones up to [5b96512](https://github.com/AstarVienna/irdb/pull/308/commits/5b96512ebf36a5c34a024b57aa39497be5a815b6) ("New notebook...")

teutoburg (approved):
Fixed the rebase accident as discussed through other means of communication. Ready to merge for me.

janusbrink (changes requested):
Thanks - this looks great. The pipeline relies on slit names "Long/Short/Wide" (and their pinhole/shifted variants) - so best to change the files/names to match before merging this change?

PR #311, opened 2026-02-05
Refactoring and formatting in tests

teutoburg:
Mostly MICADO. Get rid of some warnings. Remove xfail from passing tests.

### DIFF
--- a/MICADO/test_micado/test_micado_imaging.py
+++ b/MICADO/test_micado/test_micado_imaging.py
@@ -1,4 +1,4 @@
-
+# -*- coding: utf-8 -*-
 """
 Tests the MCAO limiting magnitudes against the values from from Ric's excel doc
 [Signal_noise_estimator_MICADO_2018.04.03]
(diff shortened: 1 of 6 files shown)


MICADO/test_micado/test_micado_imaging.py
```diff
@@ -62,19 +61,35 @@ class TestLimiting:
...
+                    reason="something changed in ScopeSim..."
+                    ),
+                ),
+            pytest.param(
+                "open", "H", 2, 27.5, 0.3,
+                marks=pytest.mark.xfail(
+                    reason="something changed in ScopeSim..."
+                    ),
+                ),
+            ("open", "Ks", 2, 27.1, 0.3),
```
hugobuddel, line 79: This is pretty ugly! But that's fine, because the uglyness is a good reminder we should fix something. And nice that this works at all.

hugobuddel (approved):

MICADO/test_micado/test_micado_imaging.py
teutoburg: Yeah, that was done by `black`, although I already reduced it somewhat, it originally put every parameter on a separate line. That's one of the cases where `black` is perhaps a bit ... overenthusiastic.

PR #323, opened 2026-03-04
Add validation test results to docs

teutoburg:
- Add a new pytest marker `validation` (see PR label).
- Refactor MICADO limiting magnitude test and add more mode combinations
- Update tolerances and xfail messages
- Configure RTD build to run pytest on (only) the validation tests (currently 1)
- Add a simple xml parser to read test results on RTD
- Add a custom Sphinx extension to render test results in table

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -5,9 +5,14 @@
 version: 2
 
 build:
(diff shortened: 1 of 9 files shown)


astronomyk (approved):
All good 👍

PR #326, opened 2026-03-12
Add PRNU effect definition to all METIS detector descriptions   

### LINKED ISSUE #325: Add description for PRNU into METIS instrument packages
Complementary to the new `PixelResponseNonUniformity` effect, a description is added in the METIS detector .YAML files: 
Preliminary values are added for the deviation. 

Example:
'- name: prnu
    description: Pixel-to-pixel response non-uniformity — fixed multiplicative gain map
    class: PixelResponseNonUniformity
    include: True
    kwargs:
      prnu_std: 0.5  
      prnu_seed: 42'

ShannonS00:
- Added the description of the PRNU variation in the METIS detector YAMLS  (IFU, IFU_SMPL, IMG_LM, Aquarius, GeoSnap)
- H2RG detectors (IFU, IFU_SMPL, IMG_LM): `prnu_std = 0.005` 
- Aquarius and GeoSnap: `prnu_std = 0.001` 
- Fixed seed (`42`) used instead of `SIM.random.seed`

Closes #325

### DIFF
--- a/METIS/METIS_DET_IFU.yaml
+++ b/METIS/METIS_DET_IFU.yaml
@@ -69,6 +69,14 @@ effects:
     description: Summing up sky signal for all DITs and NDITs
     class: ExposureIntegration
 
+  - name: prnu
+    description: Pixel-to-pixel response non-uniformity — fixed multiplicative gain map
+    class: PixelResponseNonUniformity
+    include: True
+    kwargs:
+      prnu_std: 0.5         # 0.5% RMS — typical H2RG value
+      prnu_seed: 42
+
   - name: dark_current
     description: METIS LMS dark current
     class: DarkCurrent
--- a/METIS/METIS_DET_IFU.yaml
+++ b/METIS/METIS_DET_IFU.yaml
@@ -17,7 +17,7 @@ properties:
   dit: "!OBS.dit"
   ndit: "!OBS.ndit"
   mindit: 1.3               # seconds, Roy van Boekel, pers. communication
-  include_prnu: False      
+  include_pixel_response: False      
   full_well: 1.e+5   # electrons, E-TNT-MPIA-1004, v1-0
   gain:
     1: 2.0            # electrons/ADU, email B.Serra 2024-10-09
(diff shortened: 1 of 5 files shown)


oczoske:
The tests currently fail because the PRs here and in Scopesim (https://github.com/AstarVienna/ScopeSim/pull/886)  both come from forked repositories and our workflow is not set up for this, so the tests use Scopesim `main` which doesn't know the effect. The tests should pass once the Scopesim PR has been merged.

It is debatable whether this effect should be included by default. If we set `include=False` then it can be included programmatically, but it would be better to implement a parameter `prnu_include` either in this yaml (as `!DET`) or in `default.yaml` (as `!OBS`) on a per-mode basis (sky modes False, wcu modes True?).

ShannonS00:
> The tests currently fail because the PRs here and in Scopesim ([AstarVienna/ScopeSim#886](https://github.com/AstarVienna/ScopeSim/pull/886)) both come from forked repositories and our workflow is not set up for this, so the tests use Scopesim `main` which doesn't know the effect. The tests should pass once the Scopesim PR has been merged.
> 
> It is debatable whether this effect should be included by default. If we set `include=False` then it can be included programmatically, but it would be better to implement a parameter `prnu_include` either in this yaml (as `!DET`) or in `default.yaml` (as `!OBS`) on a per-mode basis (sky modes False, wcu modes True?).


--> Ok! I think I would add it as a parameter in as `!DET` in the YAMLS

METIS/METIS_DET_IFU_SMPL.yaml
oczoske, line 1: I think we should keep prnu out of IFU_SMPL entirely - this is a mode for scientists, that simulates the (pipeline-processed!) result of a sequence of exposures and doesn't even produce a proper detector image. The effect of prnu on this simulations would need a bit more thought but is negligible anyway.

METIS/METIS_DET_IFU.yaml
```diff
@@ -69,6 +70,14 @@ effects:
     description: Summing up sky signal for all DITs and NDITs
     class: ExposureIntegration
 
+  - name: prnu
```
oczoske: Let's call it `pixel_response` - less cryptic...

METIS/METIS_DET_IMG_LM.yaml
```diff
@@ -82,6 +83,14 @@ effects:
     description: Summing up sky signal for all DITs and NDITs
     class: ExposureIntegration
 
+  - name: prnu
+    description: Pixel-to-pixel response non-uniformity — fixed multiplicative gain map
```
oczoske: Please, no en-dashes...

oczoske (changes requested):
Sorry, the review was "pending" and I wasn't aware that nobody could see my comments...

METIS/METIS_DET_IFU.yaml
ShannonS00: Ok, I would then also change the `include_prnu` to  `include_pixel_response`

ShannonS00: ... or leave it?

oczoske (approved):

PR #331, opened 2026-04-02
Fix some tests after `system_transmission` was converted to method 

oczoske:
- HAWKI, OSIRIS and WFC3 tests treated `sim.OpticsManager.system_transmission` as a property, which caused tests to fail after https://github.com/AstarVienna/ScopeSim/pull/894.
- Removed a pdf from the HAWKI package, which made up 97% of the package size (almost 10Mb)
- Bit of linting...


teutoburg (approved):
LGTM 👍
I removed the `instrument-specific` label, because this concerns three packages.
Thanks for the cleanup / linting!

PR #332, opened 2026-04-07
MICADO IFU mode

oczoske:
A humble start for now.


oczoske:
The branch has been merged into `oc/asymm_slit` and is thus part of #343 .

PR #336, opened 2026-04-17
Add illumination effect inclusion

ShannonS00:
- Include the Illumination effect as `"!INST."` , default is set to `FALSE`

- name: illumination
description: Large-scale illumination variation across the image plane
class: Illumination
include: "!INST.include_illumination"


oczoske (approved):
I'd like to have ExtraFitsKeywords last, but we can do that at some later stage.

PR #337, opened 2026-04-17
Bump `needs_scopesim` in preparation of release

teutoburg:
MICADO needs 0.11.3 because of offset slits (only spec mode but reasonable to bump top-level as well for release).
METIS needs 0.11.3 because new effects.

### DIFF
--- a/METIS/default.yaml
+++ b/METIS/default.yaml
@@ -6,7 +6,7 @@ alias: OBS
 name: METIS_default_configuration
 description: default parameters needed for a METIS simulation
 status: development
-needs_scopesim: "v0.11.1a1"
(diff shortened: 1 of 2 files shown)


oczoske (changes requested):
The MICADO slit wheel has offset slits, those need 0.11.3. The notebook that demonstrates has been moved out of the testing way, so that is not immediately apparent. I think we can put the through-slit imaging into a manageable notebook that can be tested.

teutoburg:
Ah, thanks for reminding me! Maybe we need `demo_notebooks` in MICADO like we have in METIS 😀

oczoske (approved):

PR #342, opened 2026-05-03
Blocking filter for MICADO

oczoske:
The blocking filter in `filter_wheel_2` in MICADO had a throughput of 0. As we know from METIS, there needs to be a finite throughput window so that the minimum and maximum wavelengths of a wave band can be determined. As for METIS, the MICADO blocking filter now has a throughput of 1e-32 over a wide wavelength range. 
To enable that, `minimum_throughput` in the effect definition had to be lowered below that level. Not sure anymore what that parameter was needed for, the scopesim code (in `ter_curve.FilterCurve`) seems rather hacky. In METIS it is set to 0.

Closes https://github.com/AstarVienna/ScopeSim/issues/917

### DIFF
--- a/MICADO/MICADO.yaml
+++ b/MICADO/MICADO.yaml
@@ -91,7 +91,7 @@ effects:
             - xY1
         filename_format: "!INST.filter_file_format"
         current_filter: "!OBS.filter_name_fw2"
-        minimum_throughput: 1.01e-4
+        minimum_throughput: 1.01e-33
         outer: 0.2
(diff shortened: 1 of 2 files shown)


teutoburg (approved):
Don't we have a Shutter effect for such dishonorable work as blocking all light? But ok, if it works...

oczoske:
We do have a `Shutter` effect, but the (IR) instruments don't have actual shutters...

PR #349, opened 2026-05-20
Change PSF filename to use variable reference

### LINKED ISSUE #350: MORFEO.yaml - change PSF filename to !RO.psf_file to match MICADO SCAO
```
-   name : morfeo_generic_psf
    description : MORFEO field varying MCAO PSF
    class : FieldConstantPSF
    kwargs:
        filename : PSF_MCAO_ConstPSF_40_18_6.fits
        warning : "Default PSF is not Field Varying. See Documentation"
```

Let's update this such that it matches the MICADO SCAO syntax for specifying external PSF files, so that it works via:

```
micado = sim.Simulation("MICADO", mode=["MCAO", "IMG_1.5mas"], 
                        properties={"!RO.psf_file": "tiptop_result_5as_K.fits"})
```

astronomyk:
Updated PSF file reference to use variable instead of hardcoded filename.

### DIFF
--- a/MORFEO/MORFEO.yaml
+++ b/MORFEO/MORFEO.yaml
@@ -6,7 +6,7 @@ description : MORFEO AO relay module
 
 properties :
     temperature : "!ATMO.temperature"
-
+    psf_file : "PSF_MCAO_ConstPSF_40_18_6.fits"
 
 effects :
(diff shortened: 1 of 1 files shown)


oczoske (approved):
Makes sense.

PR #357, opened 2026-05-29
Add missing TECH keywords to most modes

### LINKED ISSUE #356: Add TECH keyword to every mode
METIS currently sets a default of `tech: "IMAGE,LM"` in the top-level OBS properties. This is overridden in e.g. `FITS_lss_keywords.yaml` to `TECH:   "LSS"`, so the headers already get the correct value. MICADO also sets a top-level default of `tech: IMAGE`, which is overridden in the new `SPEC` mode to `tech: SPEC`, which can then be looked up via `!OBS.tech`. I propose to set a `tech` in _every_ mode's `properties` and remove the top-level default. The FITS header yamls should then just reference this, as is already done in MICADO in `FITS_extra_keywords.yaml`: `TECH: "!OBS.tech"`. This would also allow other tools (e.g. `scopesim-service`) to efficiently query the modes in the default yamls.

One thing that needs to be done before making this otherwise trivial change is to find out from the reference docs what exactly the value should be for each mode in METIS and MICADO.

Related to #240 and #329.

teutoburg:
Using best-guess values for MICADO, for METIS taken from: https://github.com/AstarVienna/irdb/issues/356#issuecomment-4553690985

Closes #356, except the LMS extended mode, which is not yet implemented anyway (also set the status to concept for that).

Not sure if some of the duplication mentioned in the comments is necessary. Probably not, but don't have time to check.
Also added keywords to the deprecated individual MICADO SPEC modes, doesn't hurt.

### DIFF
--- a/METIS/default.yaml
+++ b/METIS/default.yaml
@@ -7,7 +7,7 @@ name: METIS_default_configuration
 description: default parameters needed for a METIS simulation
 status: development
 needs_scopesim: "v0.11.3"
-date_modified: 2025-10-09
+date_modified: 2026-05-29
 changes:
   - 2021-12-16 (OC) chopnod defaults to perpendicular
   - 2021-12-16 (OC) default slits renamed
@@ -23,6 +23,7 @@ changes:
   - 2025-09-29 (OC) version bump
   - 2025-10-09 (OC) proper MJD value
   - 2026-03-31 (OC) set wave_min to 2.65um for bluest LMS settings
+  - 2026-05-29 (FH) added tech keyword to most modes
 
 packages:
   - Armazones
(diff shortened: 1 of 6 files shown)


hugobuddel (approved):
I still think we should go for a 'mode' per observing template, and settle these header keywords once and for all.  This is a step into that direction, so approved.

oczoske (commented):
This is okay. It doesn't really change anything for METIS except reduce the hard-coding and making it easier for thinks like METIS_Simulations to change the keyword values, should  the need arise. I would prefer it if the keyword were called `!OBS.dpr_tech` -- an easy fix for METIS, but  for MICADO `catg` and `type` would have to be renamed similarly. 
While we're at it, could the wcu modes get `dpr_catg = CALIBRATION` and the standard modes `SCIENCE`  or are there cases where that is not appropriate? Standard stars?

PR #358, opened 2026-06-15
Demo notebook for SkycalcTERCurve

oczoske:
The notebook shows how to work with `SkycalcTERCurve`, following https://github.com/AstarVienna/ScopeSim/pull/853.
It is placed among the other demo notebooks for METIS for the time being (to keep them in one place), although the effect is more generally used.


teutoburg (approved):
I don't have time to read through it right now, but it's committed without output, so that's fine.

PR #365, opened 2026-07-28
MICADO calibration mode

### LINKED ISSUE #363: MCA calibration unit support for MICADO
A placeholder issue to capture the requirements for adding the ScopeSim / IRDB support for flux-calibrated simulations with the MICADO calibration system.

oczoske:
This PR implements a submode using the MICADO calibration assembly. ELT and Armazones atmosphere are not included in this mode. The PSF is only instrumental, and the absorption spectrum of the air in the path through the relay optics is taken into account. The mode stands in for the AO modes and can be combined with all the instrument modes, e.g.
```
cmd = sim.UserCommands(use_instrument="MICADO", set_modes=["CALIB", "IMG_1.5mas"])
```
The mode should be used with a `Source` (to be defined).

Closes #363


janusbrink (approved):
Thanks - this looks great! We should remember to confirm the default PSF size with JU.
Regarding the relay optics. We need to add two surfaces specific to the MCA: the deployable mirror and the "filter" that represents the air column. Also, later the MCA would be used with MORFEO, so then we should replace the RO and AO mode definitions used for calibration appropriately.

oczoske:
The air transmission in the relay optics looks like this (in effect `air_transmission`, available in steps of 5% in relative humidity, only subset plotted here):
MICADO_water_trans

MICADO/change.log
teutoburg, line 1: Wow, how did that survive until now?

janusbrink (approved):
This looks fine to me. Two questions, though
* I do not see the actual `MICADO_water_trans.fits` file in the IRDB yet?
* ATMO.temperature is referenced in `MICADO_Standalone_RO.yaml`, but I'm not sure it will have a default value (or is used) in this mode?

teutoburg (approved):

PR #369, opened 2026-08-11
Move laser parameters to config yaml

oczoske:
- Laser wavelengths and powers are now in `metis_wcu_config.yaml` instead of hard-coded in ScopeSim.  
- N-band lasers have been added with wavelengths 8.0, 9.0, 10.4, 11.5, 12.4 um and powers of 2.0, 1.4, 2.0, 0.8, 0.5 mW, respectively. All N-band lasers are on at the same time. 
- A section has been added to `METIS_WCU.ipynb` to explain how to change the settings of the tunable laser.

### DIFF
--- a/METIS/metis_wcu_config.yaml
+++ b/METIS/metis_wcu_config.yaml
@@ -15,6 +15,7 @@ fpmask_shift: [0, 0]
 laser_l_wave: 3.39    # [um]
 laser_l_power: 5e-3   # [W]
 laser_t_wave: [4.68, 4.69, 4.70, 4.71, 4.72, 4.73, 4.74, 4.75, 4.76, 4.77]  # [um], array
+laser_t_wave_limits: [1., 30.]  # um, for validation
 laser_t_power: 70e-3  # [W], will be divided by number of lines
 laser_m_wave: 5.26    # [um]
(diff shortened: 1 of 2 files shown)


teutoburg (approved):

METIS/metis_wcu_config.yaml
```diff
@@ -11,6 +11,15 @@ current_fpmask: "open"
 fpmask_angle: 0
 fpmask_shift: [0, 0]
 
+# ------------- Available lasers
+laser_l_wave: 3.39    # [um]
+laser_l_power: 5e-3   # [W]
+laser_t_wave: [4.70]  # [um], array
```
janusbrink: If we could keep the default tuneable laser line list as it was in ScopeSim that would help:
`laser_t_wave: [4.68 4.69 4.7  4.71 4.72 4.73 4.74 4.75 4.76 4.77 4.78]`

janusbrink (changes requested):

METIS/metis_wcu_config.yaml
oczoske: I've made it a round 10 lines, so 4.77 is the bluest, if that's alright.

teutoburg (approved):

PR #371, opened 2026-08-17
Convert and fix METIS demo notebooks

teutoburg:
Converted to MyST markdown format for reasons discsussed elsewhere. Fixed headings (see 2nd commit) and some (very) small fixes. Moved the `link_irdb` line to a separate setup cell that's injected during testing, needs to be wired into the RTD build somehow, but that's currently broken anyway. From experience, people just **do not read** instructions like "uncomment this line if ...", it's annoying but we need to deal with it.

To view these notebooks as a normal Jupyter notebook, install `jupytext` and run `jupytext-config set-default-viewer markdown`, then they should open and run just like `.ipynb` files. Editing and saving them should also work.

See AstarVienna/scopesim-targets#166 for explanation on the MyST NB test script.


hugobuddel (approved):
I'm assuming that people can still run the MyST notebooks as a jupyter notebook right?  I can't easily test that now.

Can you also change ScopeSim_Data?

I still think the `conftest.py` is way too complex, but well.

teutoburg:
> I'm assuming that people can still run the MyST notebooks as a jupyter notebook right?  I can't easily test that now.

Yes, and I'm going to note _how_ above, hence the "More description will follow". It's trivial in principle, but I need to find it again. I didn't intend to request your review before adding that part. I undrafted to see if the tests would run correctly, forgetting that (a) this will trigger an automatic review request (of the team in general and you specifically, because it touches .github), and (b) I could have run the notebook tests manually for the branch. Sorry

teutoburg:
> Can you also change ScopeSim_Data?

Sure, shouldn't be too hard.

Updated the PR description about viewing these in normal Jupyter.

oczoske (approved):

PR #374, opened 2026-09-10
Add flag keywords for MICADO detector window

teutoburg:
These allow the "full detector array" mode to be a "first-order- configurable", i.e. can be set via `properties`, instead of having to switch the `.include` on the two effects after instantiating the optical train. The latter still works, so this is fully backwards compatible.


oczoske (approved):
Mostly harmless

PR #375, opened 2026-09-17
Include dark current in AutoExposure

oczoske:
Write `dark_current` explicitely in `AutoExposure` properties to enable https://github.com/AstarVienna/ScopeSim/pull/1031


teutoburg (approved):
When are we going to add an `AutoExposure` effect to MICADO?

PR #378, opened 2026-09-21
Notebook updates after test against Scopesim 0.12.0rc1

oczoske:
Main change is the increase in the offset of the MICADO offset slits: `MICADO_spectro_demo.ipynb` showed that the offset must be around 3 arcsec to fill the slit gap (actual value is unknown, although the slits apparently exist). The remaining notebooks feature updates to the current interface (the generic ``SPEC`` mode in MICADO) and cosmetics.


METIS/docs/example_notebooks/Introduction_to_Scopesim_for_METIS.pdf
teutoburg, line 1: PDF? meh...

teutoburg (approved):
Thanks a lot for dealing with all of this, especially the consistent headings!

